Issue #37, opened 2017-03-30
When VM RAM amount is less than 1GB display it in MB


bwsw:
E.g. 0.5 GB > 512 MB, because the user expects to see it in that way.

PR #68, opened 2017-04-17
#37 show RAM in MB if it is less than 1 GB

andrewbents:


src/app/vm/vm-list/vm-list-item.component.html
```diff
@@ -40,7 +40,7 @@
       </thead>
       <tr>
         <td>{{ vm.cpuNumber + 'x' + vm.cpuSpeed }} {{ 'MHZ' | translate }}</td>
-        <td>{{ vm.memory | division:2:10:2 }} {{ 'GB' | translate }}</td>
+        <td>{{ memory | async }}</td>
```
zolotyx: Please find a way to use translatePipe instead of translateService here.

src/app/vm/vm-list/vm-list-item.component.html
```diff
@@ -40,7 +40,7 @@
       </thead>
       <tr>
         <td>{{ vm.cpuNumber + 'x' + vm.cpuSpeed }} {{ 'MHZ' | translate }}</td>
-        <td>{{ vm.memory | division:2:10:2 }} {{ 'GB' | translate }}</td>
```
zolotyx: Please find a way to use translatePipe instead of translateService here.

PR #70, opened 2017-04-17
Floating point number restriction for security group rules

wowshakhov:


src/app/shared/directives/integer-validator.ts
```diff
@@ -0,0 +1,16 @@
...
+
+export function integerValidator(): ValidatorFn {
+  return (c: AbstractControl) => {
+    const isValid = Number.isInteger(+c.value);
+
+    if (isValid) {
+      return null;
+    }
+    return {
+      control: {
```
zolotyx: Please use correct validation name here i.e. 'integer' or 'validateInteger'

PR #88, opened 2017-04-21
Tagstorage 78

wowshakhov:


src/app/shared/services/storage.service.ts
```diff
@@ -1,4 +1,6 @@
 import { Injectable } from '@angular/core';
```
zolotyx: Please add unit tests for StorageService

src/app/shared/services/language.service.ts
```diff
@@ -1,6 +1,8 @@
 import { Injectable } from '@angular/core';
 import { StorageService } from './storage.service';
 import { TranslateService } from '@ngx-translate/core';
+import { Observable } from 'rxjs';
```
andrewbents: don't import the whole rxjs, check other files

src/app/app.component.html
```diff
@@ -20,9 +20,9 @@
         <div class="link-element">{{ 'WELCOME' | translate:{ name: title } }}</div>
       </div>
     </mdl-layout-title>
-    <nav class="mdl-navigation">
-
+    <nav #navigationBar class="mdl-navigation">
       <a
+        #instancesLink
```
andrewbents: remove the ids on the anchors

PR #93, opened 2017-04-25
Vmnames 30

wowshakhov:


src/app/vm/shared/vm.service.ts
```diff
@@ -358,4 +379,10 @@ export class VmService extends BaseBackendService<VirtualMachine> {
     });
     return vm;
   }
+
+  private writeNumberOfVms(number: number): Observable<number> {
+    return this.userService
+      .writeTag('numberOfVms', number.toString())
+      .map(() => number);
```
andrewbents: use mapTo

src/app/vm/shared/vm.service.ts
```diff
@@ -54,11 +55,31 @@ export class VmService extends BaseBackendService<VirtualMachine> {
...
+    private volumeService: VolumeService
   ) {
     super();
   }
 
+  public getNumberOfVms(): Observable<number> {
+    return this.userService.readTag('numberOfVms')
+      .switchMap(numberOfVms => {
+        if (numberOfVms !== undefined) {
+          return Observable.of(+numberOfVms);
```
andrewbents: it may be possible that numberOfVms wouldn't convert to a number

PR #105, opened 2017-04-28
#6 add first day of the week option

### LINKED ISSUE #6: Customize calendar control to support two types of weeks Sunday - Saturday or Monday - Sunday
Actually there are two types of them as specified in the title. We have to add to the settings the chooser which allows to switch the user between them.

andrewbents:


wowshakhov:
chech if it's possible to improve the behavior of selects w/ translations

Issue #31, opened 2017-03-29
Description and OS in Images area / Sidebar


bwsw:
1. Remove "Description" label from common info bar in the sidebar and allow description to be multiline. With current model "Description: <actual description>" it seems that there is not enough of space to display long descriptions properly

So the template:
CentOS 7 (64-bit) no GUI
Description: Image description

Will look like:
CentOS 7 (64-bit) no GUI
Image description

2. Also, if the solution implemented, then "OS" label space must be shorter to allow "OS" value fitting well in most cases (just try to determine which is longest in current CS 4.9.2) or use overflow hidden if it doesn't fit with a tooltip to see full.

bwsw:
Also, if the solution implemented, then "OS" label space must be shorter to allow "OS" value fitting well in most cases (just try to determine which is longest in current CS 4.9.2) or use overflow hidden if it doesn't fit with a tooltip to see full.

PR #108, opened 2017-05-02
#31 remove description label in isos, add multiline description

wowshakhov:


src/app/template/template/template.component.scss
```diff
@@ -8,14 +8,19 @@
   }
 }
 
+.last {
+  flex: 6 !important;
```
andrewbents: remove flex from 'last' and 'first' and use margin

PR #122, opened 2017-05-04
#8 Eventfiltering

wowshakhov:


src/app/events/event-list.component.spec.ts
```diff
@@ -0,0 +1,229 @@
...
+      .createComponent(EventListComponent);
+    comp = fixture.componentInstance;
+  });
+
+  it('should grab events by date', () => {
+    fixture.detectChanges();
+    comp.date = new Date('1970-01-02');
+    comp.getEvents();
+    fixture.detectChanges();
+    expect(comp.tableModel.data[0].id).toBe('1');
```
andrewbents: I think you should also check that the view renders tableModel correctly

src/app/events/event-list.component.ts
```diff
@@ -19,76 +21,154 @@ export class EventListComponent implements OnInit {
...
+  }
+
+  private stringifyDate(date: Date): string {
+    const options = {
+      hour: 'numeric',
+      minute: 'numeric',
+      second: 'numeric',
+      timeZoneName: 'short'
+    };
+    const dateTimeFormat = new Intl.DateTimeFormat(this.locale, options);
```
andrewbents: avoid constructing new formatter for every event

Issue #121, opened 2017-05-03
Add to account settings session feature to allow user set inactivity timeout in minutes


bwsw:
The UI must prevent from API token timeout until this inactivity occured. So, basically, request for account information periodically (every 1  minute) to avoide API timeout, but if there is no activity in Timeout minutes, then logout automatically.

PR #133, opened 2017-05-11
#121 Inactivity

wowshakhov:


src/app/shared/services/http.service.ts
```diff
@@ -0,0 +1,6 @@
+import { Injectable } from '@angular/core';
+import { Http } from '@angular/http';
+
+
+@Injectable()
+export class HttpService extends Http {}
```
zolotyx: Do we need this?

src/app/shared/services/auth.service.ts
```diff
@@ -1,12 +1,16 @@
...
-import { ErrorService } from './error.service';
-import { BackendResource } from '../decorators/backend-resource.decorator';
+import { BaseModelStub } from '../models';
+import { BackendResource } from '../decorators';
+import { StorageService } from './storage.service';
+import { UserService } from './user.service';
+
+import debounce = require('lodash/debounce');
+
+export const SESSION_REFRESH_INTERVAL = 1000;
```
zolotyx: Please move it to config

Issue #24, opened 2017-03-29
Add API section in Settings


ershovaelena:
It would be great to create an API section in Settings.
Display:

- API URL, 
- Key, 
- Token, 
- Link to API documentation

PR #149, opened 2017-05-17
#24 add api section to settings

wowshakhov:


src/app/settings/api-info/api-info.component.html
```diff
@@ -0,0 +1,45 @@
+<div class="settings-control">
+  <h4 [translate]="'API'"></h4>
+
+  <div *ngFor="let l of linkFields">
```
zolotyx: Let's remove this loop.
Just insert required fields explicitly

src/app/settings/api-info/api-info.component.ts
```diff
@@ -0,0 +1,62 @@
...
+
+        this.inputFields = [
+          { title: 'API_KEY', value: apiKey },
+          { title: 'API_SECRET_KEY', value: secretKey }
+        ];
+      });
+  }
+
+  private get apiUrl(): string {
+    return location.origin + BACKEND_API_URL;
```
zolotyx: Do all the browsers support localtion.origin?
may be we need to add this?
` if (!window.location.origin) {
        window.location.origin = '' +
            window.location.protocol + '//' +
            window.location.hostname +
            (window.location.port ? ':' + window.location.port : '');
    }`

src/app/shared/services/user.service.ts
```diff
@@ -22,6 +22,28 @@ export class UserService extends BaseBackendService<User> {
     super();
   }
 
+  public getApiKey(): Observable<string> {
+    return this.getList()
```
zolotyx: Should we try to fetch single user instead of the whole list?
We can also use getList directly in the component and try to get user.apiKey there.
It can save us a single call

src/app/settings/api-info/api-info.component.ts
```diff
@@ -0,0 +1,101 @@
...
+          apiKey: {title: 'API_KEY', value: apiKeys.apiKey},
+          apiSecretKey: {title: 'API_SECRET_KEY', value: apiKeys.secretKey}
+        };
+      });
+  }
+
+  private get apiUrl(): string {
+    let locationOrigin: string;
+
+    if (location.origin) {
```
zolotyx: Let's create utils method for this

zolotyx: Let's create utils method for this

PR #185, opened 2017-06-02
(closes #168) add material colors to vm color picker

### LINKED ISSUE #168: Expand VM color picker
Add colors from system color picker (in Settings) to VM color picker, there should be bright colors, not only neutral as there are now

andrewbents:


src/app/vm/vm-list/vm-list-item.component.scss
```diff
@@ -16,6 +16,13 @@
   }
 }
 
+.custom-card.has-text-color {
```
zolotyx: Please add small box-shadow around the status indicator

src/app/vm/shared/vm.service.ts
```diff
@@ -291,10 +293,19 @@ export class VmService extends BaseBackendService<VirtualMachine> {
     if (vm.tags) {
       let colorTag = vm.tags.find(tag => tag.key === 'color');
       if (colorTag) {
-        return new Color(colorTag.value, colorTag.value);
+        const [backgroundColor, textColor] = colorTag.value.split(VmService.vmColorDelimiter);
```
zolotyx: What if we move getColor method to vm Model?

PR #211, opened 2017-06-21
(closes #183) Add support for snapshot description

### LINKED ISSUE #183: Support snapshot description if possible
If ACS snapshot API support tags, then implement the feature that enables setting arbitrary description to created snapshots (only for created with manual mode). This allows a user to make a note why that snapshot was taken.

1. Snapshot creation: put "Description" optional field after Snapshot name
2. A Description should appear in "View all" table (Storage view-> Disk snapshot section). Put description in the table - under the name, date and actions, in the same cell (see the pic). 
snap

andrewbents:


src/app/shared/services/snapshot.service.ts
```diff
@@ -29,7 +37,14 @@ export class SnapshotService extends BaseBackendCachedService<Snapshot> {
     }
 
     return this.sendCommand('create', params)
-      .switchMap(job => this.asyncJobService.queryJob(job, this.entity, this.entityModel));
+      .switchMap(job => {
+        const asyncJob = this.asyncJobService.queryJob(job, this.entity, this.entityModel);
+        const tag = description
+          ? this.tagService.update({ id: job.id }, 'Snapshot', DESCRIPTION_TAG, description)
```
zolotyx: Why not to use this.entity instead of "Snapshot"?

PR #234, opened 2017-06-27
Angular CLI integration

zolotyx:


tslint.json
```diff
@@ -13,42 +13,55 @@
...
     "indent": [
       true,
       "spaces"
     ],
+    "interface-over-type-literal": true,
     "label-position": true,
     "max-line-length": [
       true,
-      120
+      140
```
andrewbents: It's too long, maybe just set it to 100 or so

src/polyfills.ts
```diff
@@ -1,16 +1,76 @@
...
+/**
+ * Need to import at least one locale-data with intl.
+ */
+// import 'intl/locale-data/jsonp/en';
 
-  Error['stackTraceLimit'] = Infinity;
 
-  require('zone.js/dist/long-stack-trace-zone');
-}
+import 'rxjs/Rx';
```
andrewbents: We should include only operators we use

src/polyfills.ts
```diff
@@ -1,16 +1,76 @@
...
+ * Edge >= 13 on the desktop, and iOS 10 and Chrome on mobile.
+ *
+ * Learn more in https://angular.io/docs/ts/latest/guide/browser-support.html
+ */
+
+/***************************************************************************************************
+ * BROWSER POLYFILLS
+ */
+
+/** IE9, IE10 and IE11 requires all of the following polyfills. **/
```
andrewbents: we need to check what polyfills we need

package.json
```diff
@@ -38,67 +25,37 @@
...
+    "@ngx-translate/core": "^7.0.0",
+    "@ngx-translate/http-loader": "^0.1.0",
     "core-js": "^2.4.1",
-    "moment": "2.18.1",
-    "ngx-clipboard": "5.0.1",
-    "reflect-metadata": "^0.1.3",
-    "rxjs": "5.4.0",
-    "uuid": "^3.0.1",
-    "zone.js": "0.8.10"
+    "moment": "^2.18.1",
```
andrewbents: Importing moment also includes all locales. The workaround is to use context replacement plugin in webpack [https://github.com/moment/moment/issues/2517#issuecomment-185836313](url). Need to find a way to do this since angular cli doesn't expose the config

package.json
```diff
@@ -3,31 +3,18 @@
...
-    "ci": "npm run e2e && npm run test",
-    "docs": "typedoc --options typedoc.json src/app/app.component.ts",
-    "start": "npm run server",
-    "start:hmr": "npm run server -- --hot",
-    "postinstall": "npm run webdriver-update"
+    "ng": "ng",
+    "start": "ng serve --proxy-config proxy.conf.js --port 8080",
+    "build": "ng build --prod",
+    "build:aot": "ng build --prod --aot",
+    "test": "ng test --code-coverage",
```
andrewbents: is there a way to run tests in watch mode?

package.json
zolotyx: Let's investigate on this feature in the scope of another PR

PR #250, opened 2017-07-06
(closes #230) Add "Refresh statistics" button to VM statistics

### LINKED ISSUE #230: Add reload icon for vm statistics pane in a sidebar
If a user is unable to reload the statistics then it has almost no useful application.
It may be implemented the same way as in the settings "regenerate keys" is implemented for API.

wowshakhov:


src/app/vm/shared/vm.service.ts
```diff
@@ -79,7 +79,7 @@ export class VmService extends BaseBackendService<VirtualMachine> {
     this.vmUpdateObservable.next(vm);
   }
 
-  public get(id: string): Observable<VirtualMachine> {
+  public get(id: string, lite = false): Observable<VirtualMachine> {
```
zolotyx: Remove "lite" option

zolotyx: Remove "lite" option

PR #252, opened 2017-07-07
(closes #224) Improve change VM group modal dialog 

### LINKED ISSUE #224: Improve Change VM Group modal dialog
1. Add group name to remove chooser, because the user should see from where the VM is removed.
2. Set current group default selected for Choose an existing group chooser. Because the user can accidently click "Assign" and the group will be changed to a wrong one.

The second case happens when more than one group exists.
image

wowshakhov:
…


src/app/vm/vm-sidebar/instance-group-selector/instance-group-selector.component.ts
```diff
@@ -1,133 +1,62 @@
...
     this.vmService.getInstanceGroupList()
+      .finally(() => this.loading = false)
       .subscribe(groups => {
-        this.groupNames = groups
-          .map(group => group.name)
-          .filter(groupName => groupName !== this.groupName);
-
-        this.checkIfAnyGroups();
-        this.setDefaultGroup();
+        this.groupNames = groups.map(group => group.name);
```
zolotyx: Can we use async pipe to pass groups to the component?

PR #255, opened 2017-07-07
(closes #254): Bug with invisible selected filter fixed

### LINKED ISSUE #254: Event list is empty if localStorage stores type that is not available for current selection
events-filter

zolotyx:


src/app/vm/vm-list/vm-list.component.ts
```diff
@@ -52,8 +52,8 @@ export class VmListComponent implements OnInit {
   public sections: Array<VmListSection> = [];
   public subsections: Array<VmListSubsection> = [];
 
-  public vmList: Array<VirtualMachine>;
-  public visibleVmList: Array<VirtualMachine>;
+  public vmList: Array<VirtualMachine> = [];
```
andrewbents: there is `*loading="!vmList"` is the template, so it doesn't show a spinner

PR #325, opened 2017-07-28
(closes # 251): Use session storage instead of local storage for activi…

HeyRoach:
…ty log page


andrewbents:
@HeyRoach fixed the tests, had to clear session storage in beforeEach

zolotyx:
Implemented in another branch

PR #349, opened 2017-08-08
(closes #346) md menu card actions

### LINKED ISSUE #346: Use md-menu instead of mdl-popover for card actions

ksendart:
replace mdl-popover with md-menu (import new modгles)

change style for button inside menu
use mdMenuTrigger instead of MdlPopoverComponent

### DIFF
--- a/src/app/security-group/sg-template-list/sg-template-list-item.component.scss
+++ b/src/app/security-group/sg-template-list/sg-template-list-item.component.scss
@@ -16,3 +16,14 @@
 .description span {
   width: auto;
 }
+
+button {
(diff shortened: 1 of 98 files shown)


src/app/security-group/sg-template-list/sg-template-list-item.component.scss
```diff
@@ -16,3 +16,14 @@
 .description span {
   width: auto;
 }
+
+button {
```
zolotyx: Please move this code from all the card components to app.scss or create a wrapper component and move styling to that component

Issue #376, opened 2017-08-10
UI should properly redirect user from url like /login?next=%2Flogin


zolotyx:
If the user goes to a url like "/login?next=%2Flogin" then after authentication login screen is being shown.
User should be redirected to VM's list in this case

PR #377, opened 2017-08-11
fix(login): add condition for queryParams

ksendart:
#376

### DIFF
--- a/src/app/auth/login.component.ts
+++ b/src/app/auth/login.component.ts
@@ -67,7 +67,8 @@ export class LoginComponent implements OnInit {
   }
 
   private handleLogin(): void {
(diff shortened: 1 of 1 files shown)


src/app/auth/login.component.ts
```diff
@@ -67,7 +67,8 @@ export class LoginComponent implements OnInit {
   }
 
   private handleLogin(): void {
-    const next = this.route.snapshot.queryParams['next'] || '';
+    const next = this.route.snapshot.queryParams['next'] && this.route.snapshot.queryParams['next'] !== '/login' ?
```
zolotyx: Please change condition to handle "login" also

PR #381, opened 2017-08-11
(closes #365) Add actions to sidebars

### LINKED ISSUE #365: Add action menu to sidebar for entity with actions
Right now only images have actions in sidebar, need to add the same for VMs and disks

wowshakhov:


andrewbents:
maybe we should put them right from the title

PR #391, opened 2017-08-14
(closes #380) Display "No data" when server responds with empty data for a given range

### LINKED ISSUE #380: [Pulse] Display "No data" when server responds with empty data for a given range
For now, chart js uses current date for X axis when the data is empty

qqcky:
display "no data" when server responds with empty data for a given range


src/app/pulse/charts/chart-area.component.html
```diff
@@ -1,13 +1,14 @@
-<div *ngIf="!charts[0]?.datasets?.length || error" class="no-charts">
+<div *ngIf="!charts[0]?.datasets?.length || error || !charts[0]?.datasets[0]?.data.length" class="no-charts">
```
andrewbents: well, the logic inside ngIfs became complicated really quick). Maybe just extract them from the template to getters or methods?

Issue #385, opened 2017-08-12
[Pulse] Choose first parameters for scale, aggregation, shift by default if no already known


bwsw:
If there are no saved parameters exist, then use [0] as default.

PR #402, opened 2017-08-17
(closes #387) Fix the view of charts' axes

### LINKED ISSUE #387: [Pulse] When displaying pulse tab where several charts are layed out. X-axe doesn't correspond
X-axe length and X-axe labels don't match. It leads to impossibility to match data on charts.

qqcky:
Fix the view of charts' axes

X-axe length and X-axe labels don't match. It leads to impossibility to match data on charts.


qqcky:
closed in #385

PR #505, opened 2017-09-06
(closes #326) Shared security groups

### LINKED ISSUE #326: Support shared SGs (as native ACS does)
To implement additional functionality to support shared SGs.

1. Rename **Firewall templates** to **Firewall**
2. Security groups that are created from template during vm creation should be marked as **private**. Private SGs should not be displayed on the Firewall page
3. Firewall page should have switch between Templates and Shared sgs
4. During VM creation user should be able to select shared SG or create private SG from template

wowshakhov:


src/app/shared/components/security-group-builder/rule-list/security-group-rule-list.component.ts
```diff
@@ -0,0 +1,11 @@
+// import { Component } from '@angular/core';
```
zolotyx: Please remove this if used

src/app/shared/components/security-group-builder/builder/security-group-rule-builder.component.ts
```diff
@@ -0,0 +1,10 @@
+// import { Component } from '@angular/core';
```
zolotyx: Please remove this if not used

Issue #256, opened 2017-07-11
Mark snapshots as deleted if disks were deleted


ershovaelena:
Snapshots of deleted disks should be marked as deleted

PR #515, opened 2017-09-08
feat(snapshots): delete snapshots

ksendart:
#256


src/app/shared/services/snapshot.service.ts
```diff
@@ -38,6 +38,10 @@ export class SnapshotService extends BaseBackendCachedService<Snapshot> {
       });
   }
 
+  public markForRemoval(snapshot: Snapshot): Observable<Snapshot> {
+    return this.snapshotTagService.markForRemoval(snapshot) as Observable<Snapshot>;
```
andrewbents: why casting? I think this method return something like success: true. Can you check that?

src/app/shared/services/volume.service.ts
```diff
@@ -132,6 +132,7 @@ export class VolumeService extends BaseBackendService<Volume> {
   }
 
   public markForRemoval(volume: Volume): Observable< Volume> {
+    volume.snapshots.forEach((snapshot) => this.snapshotService.markForRemoval(snapshot).subscribe());
```
zolotyx: This should not work like this. We should make all the async requests. and subscribe to the observable

src/app/shared/services/volume.service.ts
```diff
@@ -132,6 +132,7 @@ export class VolumeService extends BaseBackendService<Volume> {
   }
 
   public markForRemoval(volume: Volume): Observable< Volume> {
+    volume.snapshots.forEach((snapshot) => Observable.forkJoin(this.snapshotService.markForRemoval(snapshot)).subscribe());
```
zolotyx: Please remove subscribe

src/app/shared/services/volume.service.ts
```diff
@@ -1,12 +1,12 @@
-import { Injectable } from '@angular/core';
-import { Observable } from 'rxjs/Observable';
-import { Subject } from 'rxjs/Subject';
-import { BackendResource } from '../decorators';
-import { Snapshot, Volume } from '../models';
-import { AsyncJobService } from './async-job.service';
-import { BaseBackendService } from './base-backend.service';
-import { SnapshotService } from './snapshot.service';
-import { VolumeTagService } from './tags/volume-tag.service';
+import {Injectable} from '@angular/core';
```
zolotyx: Please fix code style

PR #520, opened 2017-09-13
(closes #429) Template groups

### LINKED ISSUE #429: Add a possibility to manage Template/ISO groups
A user should have a possibility to group and filter templates by different characteristics. 

**Acceptance criteria:**
1. Set Template/ISO groups IDs (machine names) and their translations in config. 
2. VM creation -> Installation source: Add a "group" selector near OS families and types
3. Images list: add a "group" selector above the list (as in VM list)
4. Images -> Details sidebar: add a section "Group" with a possibility to change it (choose existing groups from config)
5. Images -> Template/ISO creation: add a "group" selector with no possibility to create new, only to choose existing in config. If the group is chosen, its id (machine name) must be stored in template tags.
6. If a user goes to tags and creates a new group that does not exist in config manually, it should be placed in **"general"** group. If he enters a group that exists, add it to this group.
7. If a template has no group, also put it in general -> when filtering by groups this template should be shown in a group "general"

Mockup

wowshakhov:
TODO: decide how to edit name translations.

Option 1) When editing a group, update all groups with the same name. In this case, every template stores translations independently, which causes duplication and may lead to inconsistencies.
Option 2) Store groups as separate entities. For example "csui.user.template-group.name.en".

At the moment editing is not implemented. You can only edit translations when creating a new group.


ershovaelena:
PR depreciated. TBD in #429

Issue #305, opened 2017-07-25
Please add listall=true  to all API calls


serverchief:
I logged in as admin user and i'm not seeing all VMs. 

I expect to see all VMs - even the ones deployed by my customers.

bwsw:
@serverchief 
Yes, it seems the attribute can be added to all lists without any damage. We will schedule it for the next iteration.

bwsw:
To consult with PO about the implementation. Task is not full. To add required properties.

bwsw:
- just add account filtering option for VM view, Disk view, Template view
- add vm free-form search field

ershovaelena:
@ksendart please, check API on admin's rights to delete other users' entities. If there are no rights to do so, disable delete buttons for entities of other users.

bwsw:
@zolotyx @ksendart 
Add account when group by account, because Name Surname is meaningless usually.

PR #535, opened 2017-09-18
feat(listAll): add listAll and account filter

ksendart:
#305


zolotyx:
1 - Please merge master
2 - add account names to domain names when grouping by Account
3 - fix this bug when a group is selected
selected grouping

Issue #547, opened 2017-09-18
Incorrect buttons layout at Recurring Snapshots dialog


bwsw:
![](https://box.bw-sw.com/f/00b08e3178c34e44be1e/?dl=1)

ksendart:
![2017-09-26 11-04-54](https://user-images.githubusercontent.com/23372522/30842174-9f687934-a2aa-11e7-8dc3-413c63d1ea6a.png)

ksendart:
![image](https://user-images.githubusercontent.com/23372522/30856480-b2c03426-a2e2-11e7-8546-1da684ca9bf5.png)

ksendart:
![image](https://user-images.githubusercontent.com/23372522/30856502-c78e07d4-a2e2-11e7-8202-f6d1531b025d.png)

PR #565, opened 2017-09-27
fix(recurring-snapshots): reorder components and update styles for Recurring Snapshots dialog

ksendart:
#547


zolotyx:
screen shot 2017-10-02 at 22 11 11

- Please fix validation message
- add a wrapper div for an Add button to reduce the height

zolotyx:
@ksendart , please change add button to plus icon

Issue #560, opened 2017-09-22
Improve shared SGs - Support displaying of many, not only one


bwsw:
Basically, a user can assign some SGs to VM (not only one). This is important from the perspective of shared SGs. 

Acceptance criteria:
1. All groups must be listed in appropriate order (if ordering means something - to investigate) in the sidebar
2. Every group should have a view button on the right side
3. Edition should not be supported from a sidebar: clicking on View button moves him to the view mode of the group in Firewall area (View dialog contains an edit button - #506). 
4. If he clicks EDIT, show him a dialog where a user is asked if he wishes to edit this group as it will affect other VMs.
"This group is shared. Other VMs that use this group will also be affected. Are you sure you want to edit this group in the Security Group management area?"
5. If he clicks YES, then open an edit dialog (#506), if no, then stay in View mode.
Connected to #506 

Mockup
![560](https://user-images.githubusercontent.com/26703076/31063206-314d276e-a75b-11e7-9b3b-1480e8ecbaec.jpg)

PR #589, opened 2017-10-04
(closes #506) Modify firewall rules view and editing

### LINKED ISSUE #506: Modify firewall view and editing (for SG templates and shared SGs)
Acceptance criteria:
1. Remove edit button from an action box for Security groups (both templates and shared). There should be only View and Delete buttons.
2. View modal window should have a filtering by  Ingress/Egress and TCP/UDP/ICMP. Grouping will be added later (issue #578)
2. View dialog should contain Edit button that opens another dialog to edit and create rules (it stays the same, just remove horizontal separation rule between existing lines).
3. Add routes
4. If the group that user wants to edit is Shared, then show a warning dialog before opening an editing (described in #560).

Connected to #560 

Mockup

qqcky:


zolotyx:
I clicked edit from vm sidebar and got this
this

qqcky:
will be closed with #560 (PR #590)

Issue #574, opened 2017-09-29
Add "Save VM password for all VMs" option


ershovaelena:
Create a modal dialog "Save for all VMs" on save password to avoid further questions. 
Add "Save VM password by default" section to Settings to set and unset this option

Acceptance criteria:
1. There is a SAVE button near a password
- in the "After VM creation" dialog (Epic #2) 
- in the password resetting dialog for VM
2. Clicking on SAVE button opens a dialog (over another one that was opened) with a suggestion to save a password for all VMs (see the 1st mockup attached).
3. There are two options depending on what a user clicks: 
- If NO was clicked - This dialog appears every time a user clicks a save password button.
- If YES was clicked -  A save password button disappears and passwords for VM are saved automatically.
4. After YES or NO was clicked, the dialog closes and a user gets back to the previous dialog.
5. There should be "Save VM password by default" section in Settings (Security (see the 2nd mockup attached). User should have a possibility to set and unset this option there (so the button will appear and disappear in the dialogs)

Mockup 
![574](https://user-images.githubusercontent.com/26703076/31061781-d267e90c-a74e-11e7-9c3f-ab37df4fcd09.jpg)
![574 2](https://user-images.githubusercontent.com/26703076/31061988-0168e7e0-a751-11e7-916f-b417736937f6.jpg)

Issue #572, opened 2017-09-29
Provide a capability to save password for VM to VM tags


ershovaelena:
Provide a user with capability to save password for VM to VM tags (create, reset password)

Acceptance criteria
A password save button should appear 
- in the "after VM creation" dialog (VM successfully created #2 ) near the password
- in the dialog with a new VM password after a password resetting

To delete VM password a user should go to Tags section (System tags) and delete it there in the "manual mode"

Mockup
![572](https://user-images.githubusercontent.com/26703076/31006453-eb22d832-a526-11e7-8d85-1df8aaa25a18.jpg)

PR #603, opened 2017-10-12
feat(save-vm-password):  save password once and by default

ksendart:
#574,
#572


src/app/vm/vm-actions/vm-reset-password-component/vm-reset-password.component.html
```diff
@@ -0,0 +1,13 @@
+<div>
+  <div [innerHTML]="translatedMessage | async"></div>
+  <div >
+    <b>{{ 'VM_POST_ACTION.VM_PASSWORD' | translate }}:</b> {{vm.password}}
+    <button md-button color="primary" *ngIf="showSaveButton"
+            (click)="savePassword()">{{ 'COMMON.SAVE' | translate }}</button>
```
zolotyx: need to show "Saved" text and prevent further clicks
And, may be we should show check_circle icon instead

src/app/vm/vm-actions/vm-reset-password-component/vm-reset-password.component.ts
```diff
@@ -0,0 +1,57 @@
...
+        }
+      };
+    this.userTagService.getSavePasswordForAllVms()
+      .subscribe(tag => {
+        this.showSaveButton = !tag;
+      });
+
+  }
+
+  public get translatedMessage(): Observable<string> {
```
zolotyx: is this method really needed?

Issue #575, opened 2017-09-29
Add account filtering option for Activity Log


ershovaelena:
- Add an account filtering option for Activity Log page for Admins

Connected to #305

PR #610, opened 2017-10-18
feat(activity log filter): add account filter to activity log page 

ksendart:
#575


src/app/events/redux/events.reducers.ts
```diff
@@ -18,6 +26,7 @@ export interface State extends EntityState<Event> {
     date: Date,
     selectedTypes: string[],
     selectedLevels: string[],
+    selectedAccounts: Account[],
```
zolotyx: let's store ids only

src/app/ssh-keys/containers/ssh-key-list.container.ts
```diff
@@ -91,12 +100,15 @@ export class SshKeyListContainerComponent extends WithUnsubscribe() implements O
...
     }, []);
 
-    const selectedAccounts = params['accounts'];
-
-    this.store.dispatch(new sshKeyActions.SshKeyFilterUpdate({
-      selectedAccounts,
-      selectedGroupings
-    }));
+    const selectedAccountIds = params['accounts'];
+    this.accounts$.takeUntil(this.unsubscribe$)
```
zolotyx: remove this because we will store ids only

PR #633, opened 2017-10-25
fix(vm-creation): Error in VM creation when affinity group contains c…

HeyRoach:
…ommas

### DIFF
--- a/src/app/vm/vm-creation/vm-creation.component.html
+++ b/src/app/vm/vm-creation/vm-creation.component.html
@@ -153,11 +153,12 @@
           </div>
 
           <div>
-            <h5>{{ 'VM_PAGE.VM_CREATION.AFFINITY_GROUPS' | translate }}</h5>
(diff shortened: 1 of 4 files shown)


src/app/vm/vm-creation/vm-creation.component.html
```diff
@@ -153,11 +153,12 @@ <h3 class="mat-dialog-title">{{ 'VM_PAGE.VM_CREATION.NEW_VM' | translate }}</h3>
           </div>
 
           <div>
-            <h5>{{ 'VM_PAGE.VM_CREATION.AFFINITY_GROUPS' | translate }}</h5>
+            <h5>{{ 'VM_PAGE.VM_CREATION.AFFINITY_GROUP' | translate }}</h5>
             <mat-form-field floatPlaceholder="never">
               <input
                 name="group"
                 matInput
+                pattern="[a-zA-Z]{1}[a-zA-Z0-9\-\_]{0,61}[a-zA-Z0-9]$"
```
zolotyx: Can we add a max length option to help user to understand why if they enter long affinity group name, save button is disabled?

PR #647, opened 2017-10-30
fix(volume-confirmation): Rename shared security group and changes in …

HeyRoach:
…confirmation


ershovaelena:
@ksendart yes, Russian dialog must be changed as well. @HeyRoach please complete the task

PR #649, opened 2017-10-31
fix(styles): VM group change dialog is cut off

HeyRoach:


src/style/app.scss
```diff
@@ -439,6 +439,7 @@ mat-table mat-header-cell,
 
 .mat-dialog-container {
   max-height: 95vh !important;
+  overflow: visible !important;
```
zolotyx: @HeyRoach , is it required to add this style to app.scss?

PR #663, opened 2017-11-10
fix(storage-filer) - fix filter for storage

ksendart:


src/app/volume/volume-filter/volume-filter.component.html
```diff
@@ -1,7 +1,7 @@
 <div class="options-panel filter-wrapper">
   <mat-checkbox
     name="volumeOnly"
-    [(ngModel)]="spareOnly"
+    [ngModel]="spareOnly"
```
zolotyx: does not work!

src/app/volume/volume-filter/volume-filter.component.html
```diff
@@ -10,7 +10,7 @@
     <mat-select
       multiple="true"
       [placeholder]="'VOLUME_PAGE.FILTERS.SELECT_ACCOUNTS' | translate"
-      [(ngModel)]="selectedAccountIds" (change)="onAccountChange.emit($event.value)"
+      [ngModel]="selectedAccountIds" (change)="onAccountsChange.emit($event.value)"
```
zolotyx: move to the next line

PR #708, opened 2017-11-28
fix(i18n) fix translations

qqcky:


src/app/shared/services/language.service.ts
```diff
@@ -21,12 +21,14 @@ export enum TimeFormat {
...
 @Injectable()
 export class LanguageService {
-
   constructor(
     private storage: LocalStorageService,
     private translate: TranslateService,
     private userTagService: UserTagService
   ) {
+    if (!translate.defaultLang) {
+      translate.setDefaultLang('en');
```
zolotyx: can we use DEFAULT_LANGUAGE constant?

src/app/settings/settings.component.ts
```diff
@@ -158,8 +147,8 @@ export class SettingsComponent extends WithUnsubscribe() implements OnInit {
   }
 
   private setLoading(): void {
```
zolotyx: why is this method required?

src/app/settings/settings.component.ts
```diff
@@ -108,8 +95,10 @@ export class SettingsComponent extends WithUnsubscribe() implements OnInit {
   }
 
   public changeLanguage(change: MatSelectChange): void {
-    this.languageService.setLanguage(change.value).subscribe();
-    this.setLoading();
+    this.isLoading = true;
```
zolotyx: I believe that we need a separate loading indicator for language. and other sections

PR #726, opened 2017-12-04
fix(account-actions): Remove action box in the card of account user

HeyRoach:


src/app/account/account-sidebar/account-sidebar.component.ts
```diff
@@ -10,9 +10,13 @@ import { AuthService } from '../../shared/services/auth.service';
...
 export class AccountSidebarComponent {
-  @Input() public entity: any;
+  @Input() public entity: Account;
   @Output() public onAccountChanged = new EventEmitter<Account>();
 
+  public get isSelf() {
+    return this.authService.user.username === this.entity.name;
+  }
+
   constructor(
```
ksendart: Hello, I suppose that there is one more case exist. We can have two accounts with the same usernames (admin for example) in the domain "Develop" and in any another subdomain. And we should have an ability to perform actions for the account 'admin' in subdomain

ksendart: http://docs.cloudstack.apache.org/projects/cloudstack-administration/en/4.8/accounts.html#users

Issue #721, opened 2017-12-01
Monthly schedule for snapshot created with incorrect date


Nadya913:
Url: http://cs-ui-ci.z1.netpoint-dc.com:8888/master/storage
User: develop

Steps:
1. Login by develop
2. Go to Storage, for Ready select Set up Schedule in action list
3. Fill values for Daily: 11.00 AM, random Time zone, random Stored snapshots
4. Select Monthly tab 
5. Press add
Actual result: error appears
Expected result: schedule created 
Severity: Major

Please, see the attach.

**Connected feature ID:** _storage_snap_schedule_

![schedule_snapshot2](https://user-images.githubusercontent.com/33111683/33473582-9cf2be02-d6a9-11e7-9c3f-e52049f577a7.png)
![schedule_snapshot](https://user-images.githubusercontent.com/33111683/33473583-9d14f6f2-d6a9-11e7-8030-fad66e5667c5.png)

PR #741, opened 2017-12-07
fix(snapshot-schedule): fix time set conditions

ksendart:
#721


zolotyx:
@ksendart , please fix snapshot list modal not to hide item if delete button is clicked, but only if user confirmed it

PR #749, opened 2017-12-08
fix(vm creation): change the condition to show firewall rule selector (closes #746)

### LINKED ISSUE #746: localStorageEnabled parameter should not impact vm creation
1 - Please remove networkTypeIsBasic method from zone model
2 - we should hide security groups on vm creation only if they are disabled in the user capabilities
3 - refactor Zone model to be interface

qqcky:
…in vm creation dialog

+ change model of Zones


src/app/shared/services/zone.service.ts
```diff
@@ -13,4 +13,11 @@ export class ZoneService extends BaseBackendCachedService<Zone> {
   constructor(protected http: HttpClient) {
     super(http);
   }
+
+  protected prepareModel(res, entityModel?): Zone {
+    if (entityModel) {
```
zolotyx: is this condition required?

src/app/vm/vm-creation/data/vm-creation-state.ts
```diff
@@ -69,10 +69,6 @@ export class VmCreationState {
     return this.diskOffering && this.diskOffering.isCustomized;
   }
 
-  public get showSecurityGroups(): boolean {
```
zolotyx: let's leave this but check capabilities also

PR #754, opened 2017-12-11
fix(sg-actions): Remove delete button for system firewall templates

HeyRoach:


src/app/security-group/sg-actions/sg-action.service.ts
```diff
@@ -17,7 +17,7 @@ const SecurityGroupShowRulesAction = {
   name: 'SECURITY_GROUP_PAGE.ACTION.RULES',
   command: SecurityGroupActionType.View,
   icon: 'visibility',
-  canActivate: (securityGroup: SecurityGroup) => true
+  canActivate: (securityGroup: SecurityGroup) => securityGroup.type !== SecurityGroupType.PredefinedTemplate
```
zolotyx: I believe that the changes should be done 7 lines above. :confused:

zolotyx: should we use canActivate or isHidden?

Issue #764, opened 2017-12-14
Wrong actions for non-domain admin user in Users tab of a sidebar


ershovaelena:
Steps to reproduce:
1. Login as a regular user
2. Go to Accounts section
3. Choose your account
4. Open "Users" tab

Expected result: Available actions for users: 
for current user:
- change password
- regenerate keys 

for other users in account: no actions

Actual result: all actions are available (as if user was domain admin)

**Connected feature ID:** _account_user_display_, _account_user_actionbox_

zolotyx:
@ershovaelena , are you sure that we need to hide it?
Native UI allows User to list current account's users

ershovaelena:
@zolotyx yes you're right. A user can access Users tab, but actions with users are limited to him. Please take a look at ticket description, it has been changed.

@ksendart will apply changes and resend PR to you.

ksendart:
Test on: http://cs-ui-ci.z1.netpoint-dc.com:8080/bwsw/764-users-tab
to test: account_user_display, account_user_actionbox

Nadya913:
http://cs-ui-ci.z1.netpoint-dc.com:8080/bwsw/764-users-tab
user1234  /1234
user123 / 123

Steps:
1. Login by  develop and go Accounts
2. Create new user for Account (user1234 was  created for user123 account )
3. Logout and login by new user
4. See the actions for user tab for Account
Expected result: for current user: change password, regenerate keys for other users in account: no actions
Actual result: all options for another user, no actions for current user

Please, see the attach.
![anotheruser](https://user-images.githubusercontent.com/33111683/34246047-84f83adc-e65f-11e7-912c-3f30bdb2ba35.png)
![current_user](https://user-images.githubusercontent.com/33111683/34246048-851b8258-e65f-11e7-973c-53ca5015af35.png)

ksendart:
Fixed.
![2017-12-21 15-49-26](https://user-images.githubusercontent.com/23372522/34247904-bc3e25b8-e666-11e7-928f-204b5fee44ef.png)
![2017-12-21 15-49-57](https://user-images.githubusercontent.com/23372522/34247905-bc6283d6-e666-11e7-9618-d86614bd0acb.png)
![2017-12-21 15-50-09](https://user-images.githubusercontent.com/23372522/34247906-bc847b8a-e666-11e7-8291-560303cefe34.png)

Nadya913:
[...]

PR #767, opened 2017-12-15
fix(users-tab):  users tab behavior for user account

ksendart:
#764


src/app/account/account-sidebar/account-users/account-users.component.ts
```diff
@@ -42,6 +42,10 @@ export class AccountUsersComponent {
     this.setStep(user.id)
   }
 
+  public get isAdmin(): boolean {
```
zolotyx: Please rename this getter. It's confuses

PR #771, opened 2017-12-15
fix(vm creation) fix new security group creation in the builder mode in the vm creation dialog

qqcky:


src/app/shared/components/security-group-builder/security-group-builder.component.ts
```diff
@@ -19,14 +30,25 @@ export class SecurityGroupBuilderComponent implements OnInit {
...
+    return NetworkRuleType;
+  }
+
+  public get NetworkRuleTypes() {
+    return NetworkRuleTypes;
+  }
 
   constructor(private securityGroupService: SecurityGroupService) {
-    this.items = [[], []];
+    this.securityGroups = [[], []];
```
zolotyx: Please change it to something like:

this.securityGroups = { available: [], selected: []}
do the same for rules
It will make code easier to read

PR #799, opened 2017-12-18
refactor(ssh-key-model): Get rid of BaseModel for SSH Keys model

HeyRoach:


src/testutils/mocks/model-services/entity-data.spec.ts
```diff
@@ -28,7 +28,7 @@ export class MockEntityData {
     this.isos = isos.map(json => new Iso(json));
     this.securityGroupTemplates = securityGroupTemplates.map(json => new SecurityGroup(json));
     this.serviceOfferings = serviceOfferings.map(json => new ServiceOffering(json));
-    this.sshKeyPairs = sshKeyPairs.map(json => new SSHKeyPair(json));
+    this.sshKeyPairs = sshKeyPairs.map(json => json as SSHKeyPair);
```
zolotyx: please change sshKeyPairs type from Array<Object> to Array<SSHKeyPair> and remove "as" keyword here and in the mock-ssh.service.spec.ts

Issue #712, opened 2017-11-30
Cross-action behaviour enhancements in VM list and sidebar 


ksendart:
Check UI during performing of several operations (two or more tasks at the same time) and correct the behavior.

The following operations should be managed in a special way if **a VM is running**:
1. VM start (VM action box)
2. VM stop (VM action box)
3. VM reboot (VM action box)
4. VM reinstall (VM action box)
5. VM destroy (VM action box)
6. VM password resetting (VM action box)
7. Service offering change (VM sidebar)
8. Affinity group change (VM sidebar)
9. SSH key change (VM sidebar)

When one of these actions is performing, others (among the same list) must be disabled (disable buttons). 

**Connected feature ID:** 
- _pending_
- _vm_start_ 
- _vm_stop_
- _vm_reboot_
- _vm_reinstall_
- _vm_destroy_
- _vm_password_reset_
- _vm_SO_change_
- _vm_afgroup_change_
- _vm_sshkey_change_

ksendart:
Test on: http://cs-ui-ci.z1.netpoint-dc.com:8080/ksendart/712-cross-action-behaviour
To test: vm_afgroup_change, vm_SO_change, vm_sshkey_change, vm_start, vm_stop, vm_reboot, vm_reinstall, vm_disk_snap_create, vm_disk_resize, vm_disk_snap_delete, vm_disk_attach, vm_disk_detach

Nadya913:
Between thestatuses the operations on sidebar are available.
For example, 
VM has started
Offering changed
VM has been stopped

Nadya913:
Still warning message for rebooting appears.
Steps to reproduce:

Go to stopped VM sidebar and click edit button in Service Offering section
Choose another service offering
Actual result: 'The virtual machine will be rebooted' is shown.
Expected result: there are no warning messages needed, stopped VM won't be rebooted.
![712-reboot](https://user-images.githubusercontent.com/33111683/34286343-93473c7e-e712-11e7-993a-d19842e76802.png)

PR #809, opened 2017-12-19
feat(cross-action-behavior) - disable actions in VM list and sidebar

ksendart:
#712


src/app/reducers/vm/redux/vm.effects.ts
```diff
@@ -843,26 +840,27 @@ export class VirtualMachinesEffects {
...
           id: notificationId,
           message: 'JOB_NOTIFICATIONS.VM.STOP_FAILED'
         });
+        this.update(vm, VmState.Error);
         return Observable.of(new vmActions.VMUpdateError(error));
       });
   }
 
-  private update(vm) {
+  private update(vm, state?: VmState) {
```
zolotyx: Let's make state mandatory

src/app/reducers/vm/redux/vm.effects.ts
```diff
@@ -843,26 +840,27 @@ export class VirtualMachinesEffects {
     this.update(vm);
     return this.vmService.command(vm, 'stop')
       .do((newVm) => {
-        this.jobsNotificationService.fail({
+        this.jobsNotificationService.finish({
           id: notificationId,
           message: 'JOB_NOTIFICATIONS.VM.STOP_DONE'
         });
-        return this.store.dispatch(new vmActions.UpdateVM(newVm));
+        return Observable.of(newVm);
```
zolotyx: do is a side effect of the stream, it should not return anything

src/app/reducers/vm/redux/vm.effects.ts
```diff
@@ -94,42 +94,39 @@ export class VirtualMachinesEffects {
...
+    })
+    .switchMap(changeAction => {
+      const vmState = changeAction.payload.vm.state;
+      const notificationId = this.jobsNotificationService.add(
+        'JOB_NOTIFICATIONS.VM.CHANGE_SERVICE_OFFERING_IN_PROGRESS');
 
+      return this.vmService
+        .changeServiceOffering(changeAction.payload.offering, changeAction.payload.vm)
+        .switchMap((newVm) => {
+          this.jobsNotificationService.finish({
```
zolotyx: this can be done using .do

src/app/reducers/vm/redux/vm.effects.ts
```diff
@@ -394,6 +390,7 @@ export class VirtualMachinesEffects {
                 id: notificationId,
                 message: 'JOB_NOTIFICATIONS.VM.DESTROY_FAILED'
               });
+              this.update(action.payload, VmState.Error);
```
zolotyx: I think we should pass a VM and an error to vmUpdateError to get rid of this.update calls

src/app/reducers/vm/redux/vm.effects.ts
```diff
@@ -94,42 +94,39 @@ export class VirtualMachinesEffects {
...
 
+      return this.vmService
+        .changeServiceOffering(changeAction.payload.offering, changeAction.payload.vm)
+        .switchMap((newVm) => {
+          this.jobsNotificationService.finish({
+            id: notificationId,
+            message: 'JOB_NOTIFICATIONS.VM.CHANGE_SERVICE_OFFERING_DONE'
+          });
+          if (vmState === VmState.Running) {
+            this.store.dispatch(new vmActions.UpdateVM(newVm));
```
zolotyx: Remove it. we should update vm inside start method

PR #822, opened 2017-12-21
refactor(asyncjob-model): get rid of BaseModel for AsyncJob model

HeyRoach:


src/app/shared/models/async-job.model.ts
```diff
@@ -1,52 +1,35 @@
...
-  public result: T;
-  public instanceType: string;
-  public instanceId?: string;
-  public resultType: string;
-  public cmd: string;
-
-  constructor(params) {
-    super(params);
-    this.mapCmd();
+export const mapCmd = (asyncJob) => {
```
zolotyx: looks like we need unit tests for exported functions

src/app/shared/services/async-job.service.ts
```diff
@@ -95,13 +94,13 @@ export class AsyncJobService extends BaseBackendService<AsyncJob<any>> {
           this.timerIds.filter(id => interval !== id);
         }
         observer.complete();
-        this.event.next(asyncJob);
+        this.event.next(Object.assign(asyncJob, {cmd: mapCmd(asyncJob)}));
```
zolotyx: can we use spread operator?

src/app/shared/services/async-job.service.ts
```diff
@@ -18,8 +18,7 @@ const enum JobStatus {
 
 @Injectable()
 @BackendResource({
-  entity: 'AsyncJob',
-  entityModel: AsyncJob
+  entity: 'AsyncJob'
 })
 export class AsyncJobService extends BaseBackendService<AsyncJob<any>> {
```
zolotyx: please check if we can get rid of **any** here

PR #834, opened 2017-12-21
fix(filters): add sorting by ABC for groupings and range system and custom security groups

qqcky:


src/app/shared/components/grouped-list/grouped-list.component.ts
```diff
@@ -40,7 +36,7 @@ export class GroupedListComponent implements OnChanges {
           name: groupings[this.level].name(groups[gn][0]),
           items: groups[gn]
         };
-      });
+      }).sort((group1, group2) => group1.name.localeCompare(group2.name));
```
zolotyx: please add a protected method sortGroup and override it instead of updateTree

PR #843, opened 2017-12-26
fix(notification-box-view): Fix view of notification-box

HeyRoach:


zolotyx:
Icons are not aligned properly:
1 - the first line width is larger than others
2 - left margin is different for close and done icons
3 - strange paddings applied: 33 and 11 px. please try to add some common solution without hacks

@HeyRoach 

icons-alignment

PR #844, opened 2017-12-26
feat(ssh-key): Store "SSH description" in account tags if supported

HeyRoach:
…ported


src/config/config.json
```diff
@@ -168,5 +168,6 @@
     "webShell": false,
     "pulse": false
   },
-  "templateGroups": []
+  "templateGroups": [],
+  "accountTags": true
```
zolotyx: no need to enable this key in git

src/config/config.json
```diff
@@ -168,6 +168,5 @@
     "webShell": true,
     "pulse": false
   },
-  "templateGroups": [],
-  "accountTags": true
+  "templateGroups": []
```
zolotyx: add accountTags = false here an rename it to accountTagsEnabled

Issue #777, opened 2017-12-15
Get rid of BaseModel for Account model 


HeyRoach:

ksendart:
test on: http://cs-ui-ci.z1.netpoint-dc.com:8080/ksendart/777-account-model
to test: 
account_card
account_actionbox
account_disable
account_enable
account_lock
account_delete
account_create

chewlite:
**Tested on:** `ksendart/777-account-model`

PR #889, opened 2018-01-16
refactor(account-model):  Get rid of BaseModel for Account

ksendart:
#777


src/app/shared/models/resource-limit.model.ts
```diff
@@ -20,3 +20,11 @@ export interface ResourceLimit extends BaseModelInterface {
   max: number;
   resourcetype: number;
 }
+
+export const updateLimitMax = (limit: ResourceLimit) => {
```
zolotyx: add return type

src/app/shared/models/resource-limit.model.ts
```diff
@@ -20,3 +20,11 @@ export interface ResourceLimit extends BaseModelInterface {
...
+  const fixedJson = { ...limit };
+  if (limit.max === -1) {
+    fixedJson['max'] = Infinity;
+  }
+  return fixedJson as ResourceLimit;
```
zolotyx: can we get rid of **as**

Issue #838, opened 2017-12-22
VMs expunge doesn't work if destroying multiple VMs at the same time


qqcky:
**Steps to reproduce:**

1. Go to VMs page 
2. Try to destroy and expunge multiple virtual machines one by one

**Expected result:** Machines disappear from the list
**Actual result:** All machines (sometimes only one disappears) are staying in the list with state `destroying`. They disappear only after reloading the page

Research the reasons of this behavior.

**Connected feature ID:** 
- _vm_destroy_
- _vm_expunge_

![](https://media.giphy.com/media/l0HU4CNS4QHxikLBK/giphy.gif)

ksendart:
test on: http://cs-ui-ci.z1.netpoint-dc.com:8080/ksendart/838-vm-expunge
to test: 
vm_start
vm_stop
vm_reboot
vm_reinstall
vm_destroy
vm_expunge
vm_recover

rennervo:
Tested on http://cs-ui-ci.z1.netpoint-dc.com:8080/ksendart/838-vm-expunge

PR #900, opened 2018-01-18
fix(vm-command):  Replace switchMap with flatMap for several vm actions

ksendart:
#838


src/polyfills.ts
```diff
@@ -61,22 +61,23 @@ import 'zone.js/dist/zone'; // Included with Angular CLI.
...
-import 'rxjs/add/operator/withLatestFrom';
+import 'rxjs/add/operator/takeUntil';
 import 'rxjs/add/operator/toPromise';
-// static operator imports ( e.g. Observable.merge )
-import 'rxjs/add/observable/merge';
-import 'rxjs/add/observable/throw';
-import 'rxjs/add/observable/fromEvent';
-import 'rxjs/add/observable/forkJoin';
-import 'rxjs/add/observable/combineLatest';
-import 'rxjs/add/operator/concatMap';
```
qqcky: Maybe we should revert not changed imports?

Issue #897, opened 2018-01-17
IE - errors for forbidden argument and not supporting includes method in console


Nadya913:
Url: http://cs-ui-ci.z1.netpoint-dc.com:8080/bwsw/master
Browser: IE 11
User: develop

Errors appear in Console for IE for few sections: VM, Images, Storage, SSH :
1. Forbidden argument
2. Method includes doesn't support

Steps:
1. Login by develop

Expected result: VMs are loaded, Resouce usage data appear
Actual result: Trobber for VM list, Resource usage 

Please see the attach.

![ie11.jpg](https://images.zenhubusercontent.com/59faba338a75884b9085bd93/8ccf952b-460a-4569-bbbd-a0750592d60e)

ksendart:
Hi! The first and main error was that IE does not support method includes for arrays (https://developer.mozilla.org/ru/docs/Web/JavaScript/Reference/Global_Objects/Array/includes#Browser_compatibility). I replaced this method with indexOf for arrays (more clear using), and also add polyfill (keep using includes for strings). The second error was an Invalid argument. As I found out, IE does not work correctly with attribute innerHTML. However, I did not find any valuable information or documentation, exсept several open issues on Stackoverflow. So, I updated html-files for several pages and also fixed styles and highlight for text searching.

ksendart:
test on: http://cs-ui-ci.z1.netpoint-dc.com:8080/ksendart/897-ie-errors

rennervo:
Fail: Account name on card is displayed halfway.
Browser: IE 11 (ver. 11.192.16299.0)
![account name ie](https://user-images.githubusercontent.com/26665398/35429342-441454fc-02a6-11e8-9e1d-a3e71cb827da.png)

rennervo:
Tested on ksendart/897-ie-errors

PR #907, opened 2018-01-22
fix(ie-errors): Fix several errors in IE.

ksendart:
#897


src/app/reducers/vm/redux/vm-creation.effects.ts
```diff
@@ -99,7 +107,7 @@ export class VirtualMachineCreationEffects {
           const insufficientResources = [];
 
           Object.keys(resourceUsage.available)
-            .filter(key => ['instances', 'volumes', 'cpus', 'memory', 'primaryStorage'].includes(key))
+            .filter(key => ['instances', 'volumes', 'cpus', 'memory', 'primaryStorage'].indexOf(key) !== -1)
```
zolotyx: Just a note, `indexOf(key) > -1 `saves time =)

Issue #871, opened 2018-01-10
Propose to delete Volume snapshots when deleting VM/Volume


ershovaelena:
Acceptance criteria:
1. When destroying a VM: add a checkbox "Delete snapshots?" ("Удалить снимки"?) to the dialog "Delete this VM's data drives?". 
This checkbox should appear **only when data drives have snapshots**
2. When deleting a Volume: add a dialog "Delete this volume's snapshots?" ("Удалить снимки данного диска?") appearing after volume deleting has been approved by user.
This dialog should appear **only when a volume has snapshots**
3. Document this feature in "Custom features" local doc

**Connected feature ID:**
- _vm_destroy_datadrives_snap_
- _storage_disk_delete_snap_

Mockup
![871](https://user-images.githubusercontent.com/26703076/34755373-31af4f9a-f5f6-11e7-92f6-5c73f023c03c.JPG)

ksendart:
test on: http://cs-ui-ci.z1.netpoint-dc.com:8080/ksendart/871-delete-snapshots
to test: 
vm_destroy
vm_expunge
storage_disk_delete
storage_disk_delete_snap
storage_snap_display
storage_snap_delete

Nadya913:
1. **vm_expunge**
Got error for "Failed to detach a volume" when VM was expunged. 
Steps:
1. Login and create new VM
2. For created VM on sidebar create new Volume to attach
3. Attach the Volume
4. Go to VMs and destroy VM, check Expunge

Expected result: disk is deleted
Actual result: error  "Failed to detach a volume"

2. **pending**
Count above Zero is shown. Please, see the attach. Appears in many scenarios: vm creation, vm destroying, disk deleting.
Steps:
1. Login and go to Storage
2. Create new, attach to VM
3. Make shap
4. Delete disk with shapshots
5. Expand the pending list
6. Close pending list

Expected result: no count value appears
Actual result: count value is above zero

![871- expunge](https://user-images.githubusercontent.com/33111683/35794029-957dd81c-0a86-11e8-80b8-12a47dc79144.png)
![871 - count](https://user-images.githubusercontent.com/33111683/35794032-99c03c4e-0a86-11e8-8207-0227e1ccbcdd.png)

Nadya913:
**pending**
Snapshot created but throbber is never ending. 
Steps:
1. Login and select any vm
2. Go to Storage and make 2 snapshots one by one for one disk
3. See the pending operations

[...]

PR #926, opened 2018-01-29
feat(delete-snapshots): Delete snapshots with data drives

ksendart:
#871


src/app/reducers/volumes/redux/volumes.effects.ts
```diff
@@ -220,17 +222,26 @@ export class VolumesEffects {
...
+      this.dialog.open(VolumeDeleteDialogComponent, {
+        data: !!volumes.find(volume => !!volume.snapshots.length)
+      }).afterClosed()
         .filter(res => Boolean(res))
-        .flatMap(() => volumes
-          .map((volume: Volume) => new volumeActions.DeleteVolume(volume))));
+        .flatMap((params) => {
+          return volumes
+            .map((volume: Volume) => {
+              if (params.deleteSnapshots && !!volume.snapshots.length) {
```
zolotyx: do not dispatch actions in effects

src/app/shared/actions/volume-actions/volume-actions.container.ts
```diff
@@ -39,6 +39,12 @@ export class VolumeActionsContainerComponent {
       .onErrorResumeNext()
       .filter(res => Boolean(res))
       .subscribe(() => {
+        if (!!volume.snapshots.length) {
```
zolotyx: Do we really need to show 2 different dialogs here?

src/app/shared/actions/volume-actions/volume-delete/volume-delete-dialog.component.ts
```diff
@@ -0,0 +1,29 @@
...
+
+  constructor(
+    public dialogRef: MatDialogRef<VolumeDeleteDialogComponent>,
+    @Inject(MAT_DIALOG_DATA) public hasSnapshots: boolean
+  ) { }
+
+  public confirmDestroy(): void {
+    const result = {};
+    if (this.deleteSnapshots) {
+      result['deleteSnapshots'] = true;
```
zolotyx: remove square brackets, use typing {deleteSnapshots?: boolean}

src/app/shared/actions/volume-actions/volume-actions.container.ts
ksendart: We have the mock. I think it should be discussed with PM and PO

ershovaelena: Yes, it was discussed with PO. We need 2 dialogs here (just the same that we have for deleted VMs)

Issue #671, opened 2017-11-18
Distinguish IPv4/IPv6 in Security Group rules view mode


bwsw:
**Acceptance criteria:**

1. Ingress SG rules are displayed incorrectly. Semantics should be like:

- Protocol (correct)
- ICMP or Ports (correct)
- **FROM** CIDR (incorrect - now is **TO**)

2. Distinguish IPv4 and IPv6. Correct:

Incoming TCP traffic **from** **IPv4** X.X.X.X/Y to **VM** (add the word) port range M-N
Incoming TCP traffic **from** **IPv6** X:Y::C/D to **VM** (add the word) port range M-N

Outgoing TCP traffic to **IPv4** X.X.X.X/Y to port range M-N	
Outgoing TCP traffic to **IPv6** X:Y::C/D to port range M-N	

IPv4 and V6 addresses can be easily distinguished with Regexp.

3. IPv4/IPv6 rules filtering: 
Filtering when view - add new **2 checkboxes ipv4, ipv6** on the right side of 2 existing filterings

**Connected feature ID:** 
- _firewall_rules_view_
-  _firewall_rules_view_filter_
- _vm_firewall_rules_view_

bwsw:
To move enchancements to ipv6 impl ticket

bwsw:
Might be it's better to implement new rule creation as modal dialog? It help define the context like, if i chosen incoming then show From if outgoing then To, like single screen master when context of the first field dwfines the next. Just to think about it...

ershovaelena:
A comment of 18.11.17 is moved to a separate issue #903

qqcky:
**To test:** 
- firewall_rules_view_filter
- firewall_rules_view
- firewall_rules_edit

Test on `qqcky/671-ipv6-support`.

rennervo:
Errors:
1. The system incorrectly determines the type of CIDR. All the rules that are created are defined as IPv6, even if it's IPv4. Because of this, the following errors occur:
 - Data is false in a row with a new rule. CIDR type is IPv6 instead of IPv4.
![ipv4 vs ipv6](https://user-images.githubusercontent.com/26665398/35897553-51865d22-0bf4-11e8-80eb-c5576a856790.png)
 - Items in fields "ICMP type" and "ICMP code" not matching with ICMPv4 protocol when an ICMPv4 rule is creating. In this case in these fields are items from the ICMPv6 protocol.

[...]

PR #935, opened 2018-01-30
feat(security-group): support ICMPv6 network rule creation (closes #920)

### LINKED ISSUE #920: IPv6 support in Security Group rules edit mode
**Acceptance criteria:**

1. Support entering IP addresses in both IPv4 and IPv6 formats (the 3d field). There should be a text `CIDR v4/v6` pre-entered in this field.

2. Support _ICMPv6_ correctly. Allow valid ICMPv6 types and codes.
(-1, -1) is incorrect for ICMPv6. Basically a user can specify -1,-1 for ICMP but lately, ACS firewall code raises errors and rules are skipped, so we have to filter it on client side allowing (right now) only valid ICMPv6 types and codes.

Consult this page: https://www.iana.org/assignments/icmpv6-parameters/icmpv6-parameters.xhtml

**Connected feature ID:** _firewall_rules_edit_, _vm_firewall_rules_edit_

qqcky:


src/app/security-group/sg-rules/sg-rules.component.html
```diff
@@ -45,12 +45,13 @@
                 name="cidr"
                 type="text"
                 [(ngModel)]="cidr"
```
ksendart: We do not need two-way binding if we use ngModel and ngModelChange in the same time

qqcky:
TBD in #671 (PR #957)

PR #936, opened 2018-01-30
feat(support-information): Create support-info.md

HeyRoach:


src/support/support-info.md
```diff
@@ -0,0 +1 @@
+For more information to write support@cloudstack.com
```
HeyRoach: it is just for testing.
After testing i will remove this file

src/app/shared/components/support-information/support-information.component.html
```diff
@@ -0,0 +1,14 @@
+<ng-container *ngIf="isShow">
+  <div class="support-information">
+    <h4>{{ 'ERRORS.SUPPORT_INFORMATION' | translate }}</h4>
+    <button
```
ksendart: Why do not you use <mat-expansion-panel>? (https://material.angular.io/components/expansion/overview)
I suppose that it looks better, and we do not need any CSS for it. It is just a comment, you can use your own solution :)

src/app/shared/components/support-information/support-information.component.ts
```diff
@@ -0,0 +1,57 @@
...
+      })
+      .subscribe(supportText => {
+        this.supportInformation = supportText;
+      });
+  }
+
+  public readFile() {
+    return this.getPath()
+      .switchMap(responsePath => this.http.get(responsePath))
+      .catch(() => this.http.get(this.defaultLangPath))
```
zolotyx: Please think if we need to cache that responses

src/app/shared/components/support-information/support-information.component.ts
```diff
@@ -0,0 +1,57 @@
+import { Component, OnInit } from '@angular/core';
+import { Http } from '@angular/http';
+import { Converter } from 'showdown';
+import { UserTagService } from '../../services/tags/user-tag.service';
+import { Observable } from 'rxjs/Observable';
+
+@Component({
```
zolotyx: Maybe a separate module should be created for this component and the functionality described in this ticket

src/app/support-information/support-information.component.ts
```diff
@@ -0,0 +1,74 @@
...
+        const converter = new Converter();
+        return converter.makeHtml(text);
+      })
+      .subscribe(supportText => {
+        this.supportInformation = supportText;
+      });
+  }
+
+  public readFile() {
+    const cachedRequest = this.requestCache.get({});
```
zolotyx: 1 - we need to cache responses, not requests
2 - are you sure that your code works correctly if we change current language?

zolotyx: can we change the signature of cache.get not to pass empty object if not required?

src/app/support-information/support-information.component.ts
```diff
@@ -0,0 +1,74 @@
...
+})
+export class SupportInformationComponent implements OnInit {
+  protected supportInformation: string;
+  protected defaultPath = 'support/support-info';
+  protected defaultLangPath = `${this.defaultPath}.${'md'}`;
+  protected isShow = true;
+  protected requestCache: Cache<Observable<FormattedResponse<string>>>;
+
+  constructor(
+    private http: Http,
```
zolotyx: should it be Http of HttpClient (please check agreement dialog and do the changes if required)

src/app/support-information/support-information.module.ts
```diff
@@ -0,0 +1,22 @@
+import { SupportInformationComponent } from './support-information.component';
```
zolotyx: Please rename module directory to **support**

src/app/shared/services/cache.ts
```diff
@@ -11,8 +11,8 @@ export interface CacheEntry<T> {
 export class Cache<T> implements ICache<T> {
   private cache: { [key: string]: T; } = {};
 
-  public get(params: {}): T {
-    const key = JSON.stringify(params);
+  public get(params?: {}): T {
+    const key = JSON.stringify(params || {});
```
zolotyx: you can set default value in method declaration

PR #952, opened 2018-02-01
 fix(volume-list): Fix Root/Data volumes wrong sorting

HeyRoach:


qqcky:
Will be done in PR #1065

Issue #950, opened 2018-02-01
Assign service offerings to classes in config.json


ershovaelena:
**Acceptance criteria:**

1. Move assigning SO to classes from account tags to config.json (related to #502, #498) 
2. Document this feature in Config guide and local doc "Custom features"
3. Remove account tag from wiki

**Connected feature ID:** 
- vm_SO_change
- vm_creation_so
- vm_creation_so_custom

ksendart:
http://cs-ui-ci.z1.netpoint-dc.com:8080/ksendart/950-so-classes
vm_SO_display
vm_SO_filter
vm_SO_change

Nadya913:
Tested on ksendart/950-so-classes
Tested for wiki, config guide, custom features

PR #955, opened 2018-02-02
fix(so-classes): Store SO classes in config.json

ksendart:
#950


src/app/shared/services/tags/account-tag.service.ts
```diff
@@ -47,7 +47,7 @@ export class AccountTagService implements EntityTagService {
     return `${this.keys.sshDescription}.${sshKey.fingerprint}`;
   }
 
-  public writeTag(key: string, value: string): Observable<string> {
+  public writeTag(key: string, value: string): Observable<any> {
```
zolotyx: why any?

ksendart: @zolotyx  I replaced string with any, because TagService.writeTag returns `TagService.update(entity: any, entityName: string, key: string, value: any): Observable<any>`. This method is normally used in different methods, which return Observables of different objects, not only strings.

zolotyx: @ksendart , but tag service always returns string values of the tag. Or not?

ksendart: @zolotyx  TagService.update accept `entity: any` returns newEntity - object with new `tags` attribute.

Issue #932, opened 2018-01-30
Shared SG is created from Firewall Template


rennervo:
Browser: IE (ver. 11.192.16299.0)
If mandatories is full and press Create when create Firewall Template, then system hang and Shared SG is created with input data after page reboot. 
An error appears in the console: "ERROR TypeError: Unable to get property 'id' of undefined or null reference"

Steps:
1. Log in and go to Firewall / Firewall Templates;
2. Press Add, full mandatories and press Create.
3. Reboot page (because the page is hang)

Actual rezult: New Shared SG created;
Expected rezult: New Firewall Template created.
Feature ID: firewall_sg_create

Screenshot:
![Hand Template.png](https://images.zenhubusercontent.com/5a5c77c24b5806bc2bc2be22/de445b88-7952-4fc7-9541-eaec97cdcb7b)

ksendart:
@Nadya913  @resmo  
Hi! The issue is not reproduced on master-branch. Please, can you check it again?

resmo:
@ksendart you mean /cc @rennervo

rennervo:
Hi! I checked it again. The issue is reproduced on master-branch.
Environment: OS in Virtualbox is Windows 10, browser is IE (ver. 11.192.16299.0).
Steps are the same as those presented above. Page was refreshed after a minute of waiting. Screenshots are attached.
![fwtssg1](https://user-images.githubusercontent.com/26665398/35838230-edc9aa46-0b1c-11e8-9f4d-b97aab28011f.png)
![wftssg2](https://user-images.githubusercontent.com/26665398/35838234-f1c9f36c-0b1c-11e8-8cd0-97be9aa87cdf.png)

ksendart:
test on: http://cs-ui-ci.z1.netpoint-dc.com:8080/ksendart/932-sg-create
to test: 
firewall_sg_create
firewall_rules_view
firewall_rules_edit
firewall_delete

chewlite:
@ksendart 

**_firewall_sg_create_**
If refresh the page when Template deploy before tag "csui.security-group.type : custom-template" was assigned, Shared SG is created instead of Template. 
It's not depends on browser

ksendart:
[...]

PR #981, opened 2018-02-07
fix(sg-create):  Disable cache, update SG creation

ksendart:
#932


src/app/shared/services/base-http-interceptor.ts
```diff
@@ -38,9 +36,18 @@ export class BaseHttpInterceptor implements HttpInterceptor {
...
-    }) : req;
+    const httpOptions = {
+      headers: new HttpHeaders({
+        'Cache-Control': 'no-cache',
+        'Pragma': 'no-cache',
+        'Expires': 'Sat, 01 Jan 2000 00:00:00 GMT'
+      })
+    };
+    const requestWithHeaders = req.clone(httpOptions);
+    const request = sessionKey ? requestWithHeaders.clone({
```
zolotyx: can we **clone** only once?

PR #1040, opened 2018-03-26
fix(snapshot): add confirmation dialog for delete action

tamazlykar:


src/app/snapshot/snapshots-page/snapshot-list-item/snapshot-actions/snapshot-action.container.ts
```diff
@@ -38,7 +38,12 @@ export class SnapshotActionContainerComponent {
   }
 
   public onSnapshotDelete(snapshot: Snapshot): void {
-    this.store.dispatch(new snapshotActions.DeleteSnapshot(snapshot));
+    this.dialogService.confirm({message: 'DIALOG_MESSAGES.SNAPSHOT.CONFIRM_DELETION'})
+      .subscribe(res => {
```
zolotyx: Please, think about using **filter** instead of conditioning next time

PR #1042, opened 2018-03-27
fix(tags): change requests order for update tag action

tamazlykar:


tamazlykar:
@zolotyx It's good idea, actually we need transfer all actions to effects, but better do this in some refactoring task, coz there is don't exist any actions for update tag.

PR #1044, opened 2018-03-28
fix(sg): limit the creation of negative ICMP type/code for IPv6

tamazlykar:


src/app/security-group/sg-rules/sg-rules.component.html
```diff
@@ -63,7 +63,7 @@
                 <input
                   matInput
                   maxValue="255"
-                  [minValue]="cidrIpVersion === IPVersion.ipv6 ? 0 : -1"
+                  minValue="0"
```
qqcky: It should be possible to use ICMP type -1 for IPv4 rules. The same for ICMP codes.

src/app/security-group/sg-rules/sg-rules.component.html
```diff
@@ -109,7 +109,7 @@
                   matInput
                   name="icmpCode"
                   maxValue="255"
-                  [minValue]="cidrIpVersion === IPVersion.ipv6 ? 0 : -1"
+                  minValue="0"
```
qqcky: It should be possible to use ICMP code -1 for IPv4 rules. The same for ICMP types.

src/app/security-group/sg-rules/sg-rules.component.html
tamazlykar: Please provide link that ICMP Type can have "-1" value. I can't find that info on the internet.
https://www.iana.org/assignments/icmp-parameters/icmp-parameters.xhtml

src/app/security-group/sg-rules/sg-rules.component.html
tamazlykar: Please provide link that ICMP Type can have "-1" value. I can't find that info on the internet.
https://www.iana.org/assignments/icmp-parameters/icmp-parameters.xhtml

tamazlykar:
@qqcky Okay, @bwsw explained that Cloudstack interpret -1 any value
I'll add comments for this in code

PR #1048, opened 2018-03-30
 build(ci): update travis configuration to build and push docker images

tamazlykar:
This problem appeared after we update Angular CLI to 1.7.3. The reason of build fails is high memory consumtion decribed there: angular/angular-cli#5618
I've tried to use described solutions in the link but it does not help to reduce memory consumption.
Due to low configuration of Docker Build machines we decided (with @zolotyx ) to use Travis CI for building and deploying our docker images.


tamazlykar:
@zolotyx I created a script directory and think it will be better if we transfer all out scripts to this directory. How do you think?

PR #1054, opened 2018-04-09
feat: change icon library to mdi

tamazlykar:
Library changed because of absent of some icons.
Absent icons was generated by CSS and it hard to maintain, this is the reason for this change.
Firstly, I tried to use SVG icons but it was very slow due to big icons pack and its require time for Angular Material to parse the pack and render this icons. 
Therefore, I used font icons.

Current library [Material Design Icons](https://materialdesignicons.com/)


src/style/app.scss
```diff
@@ -586,3 +579,8 @@ h3 {
   overflow-wrap: break-word;
   word-wrap: break-word;
 }
+
+.mdi {
```
vanekbr: @tamazlykar Do we need this style? You didn't add this class to any html tag. MDI library itself has default font size 24px

tamazlykar: @vanekbr Yes, we need it. Because icons implements by font, they rely on font-size attribute. Angular Material have rule .mat-icon-button that define font-size at 14 px, so because of this we need to provide our own desired size for icons.
This .mdi class applies to icons in indirect way. We can register class that will be applied to every <mat-icon> tag in application. It registered in app.module (link below), also MDI library require .mdi class for every icon to properly render icon.

https://github.com/tamazlykar/cloudstack-ui/blob/aba1ab988728d0792caa6b0ec219d81cfd343b22/src/app/app.module.ts#L143-L145

tamazlykar:
Please check the changes that I made to fix bugs found at testing.

/cc @zolotyx @vanekbr

src/app/shared/badge/badge.directive.ts
```diff
@@ -1,21 +1,14 @@
...
 export class BadgeDirective {
-  @HostBinding('class.badge') public readonly badge = true;
-  @HostBinding('attr.data-badge') @Input() public csBadge: string;
-
-  @HostBinding('class.badge--overlap') private shouldOverlap = false;
-
-  @Input()
-  public get csBadgeOverlap(): boolean {
-    return this.shouldOverlap;
+  @HostBinding('class.badge')
```
zolotyx: do we still need this directive?

src/app/shared/badge/badge.scss
```diff
@@ -14,7 +13,7 @@ $badge-overlap: 14px;
     align-items: center;
     position: absolute;
     top: -($badge-size / 2);
-    right: -($badge-size + $badge-offset);
+    right: -($badge-size - $badge-overlap);
```
zolotyx: please, move this to the component dir

tamazlykar: @zolotyx fixed

PR #1056, opened 2018-04-11
fix(account): fix the ability to delete the current account from sidebar

tamazlykar:
This is a fix for the [previous PR](https://github.com/bwsw/cloudstack-ui/pull/1043) that was merged by mistake

### DIFF
--- a/src/app/shared/utils/account.ts
+++ b/src/app/shared/utils/account.ts
@@ -1,5 +1,5 @@
 import { User, Account } from '../models';
 
-export function isHaveSameAccount(user: User, account: Account) {
-  return user.account === account.user[0].account;
+export function isUserBelongsToAccount(user: User, account: Account) {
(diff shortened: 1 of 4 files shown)


tamazlykar:
@vanekbr changes provided

src/app/shared/utils/account.ts
```diff
@@ -0,0 +1,5 @@
+import { User, Account } from '../models';
+
+export function isHaveSameAccount(user: User, account: Account) {
+  return user.account === account.user[0].account;
```
zolotyx: Please make sure that user array is not empty

tamazlykar: API doc tells that account.user implies element has a child.
Do I need to write an additional check in the code?

zolotyx: yes, and please, rename function to represent the meaning "User belongs to the account"

PR #1060, opened 2018-04-17
fix: repair group by filter

tamazlykar:
Fix the order of the "group by" filter elements after reloading the page or navigating to other sections.


src/app/account/account-list-filter/account-list-filter.component.ts
```diff
@@ -33,4 +28,7 @@ export class AccountListFilterComponent {
     return stateTranslations[state];
   }
 
+  public ngOnInit() {
+    this.groupings = reorderAvailableGroupings(this.groupings, this.selectedGroupings);
```
zolotyx: will this work if groupings are changes from outside of this component after ngOnInit was called?

tamazlykar: No, it would not, but I can't see any use cases that can change selected groupings outside.
This only works when we create a component and we don't' want behavior such as if user select groupings it will automatically go up to the top of the grouping list.

src/app/shared/utils/reorder-groupings.ts
```diff
@@ -0,0 +1,7 @@
+import { Grouping } from '../models';
+
+export function reorderAvailableGroupings(availableGroupings: Grouping[], selectedGroupings: Grouping[]): Array<any> {
```
vanekbr: Try to avoid "any" type, replace it with the exact type of returned value

src/app/account/account-list-filter/account-list-filter.component.ts
zolotyx: Ok, but if we will have some logic like  "Groupings depend on the data" For example, we want to remove groupings if no grouping is required for that particular data set, them this approach will not work. But let's proceed with this solution

Issue #1000, opened 2018-02-20
Ability to add a Firewall rule with empty ICMP type/code


chewlite:
There is no validation for empty ICMP type/code and no asterisks but these fields are required.
There is ability to add a rule when _empty ICMP type_ or _ICMP code_ or _ICMP type and ICMP code simultaneously_

Steps:

**1.** Go to Firewall rules editing
**2.** Select ICMP protocol and input valid IP

**3.1** Click on empty ICMP type/code then click on neutral area
Expected result: ICMP type/code is invalid, add-button is disabled
Actual result: ICMP type/code is valid, add-button is enabled

**3.2** Fill "ICMP type/code" by some numbers, then delete _type_ or _code_ or _two of them simultaneously_ and click add
Expected result: add-button is disabled 
Actual result: add-button is enabled, the rule created with type and code
![deleted type](https://user-images.githubusercontent.com/32011943/36419066-78e80d88-1663-11e8-8d5f-826d10037b5d.png)
![deleted code](https://user-images.githubusercontent.com/32011943/36419071-7b8bb2e2-1663-11e8-9a93-88a5ec6c753d.png)
![deleted type code](https://user-images.githubusercontent.com/32011943/36419076-7dcf3290-1663-11e8-88cb-39b8c9872454.png)

**Connected Feature ID:**
_firewall_rules_edit
vm_firewall_rules_edit_

chewlite:
Also there is ability to add the rule and no validation when Start port is greater then End port

ershovaelena:
TBD in #1019

tamazlykar:
PR fixes #1019, #1000, and 3.2 item at #997
Please test there only #1000

Test on:
tamazlykar/1019-firewall-rules

Regression:
firewall_rules_edit
vm_firewall_rules_edit

chewlite:
@tamazlykar 

Fails in **_firewall_rules_edit, vm_firewall_rules_edit_**:

1) If I understand comment by @qqcky in [#920](https://github.com/bwsw/cloudstack-ui/issues/920) correctly, unassigned types and codes also could be entered in ICMP type/code fields. Then we must have an ability to enter any value [0;255] and it must be valid

2) Steps:
1 - input valid CIDR, ICMP type and code (or don't touch ICMP type and code, or input and clear ICMP type and code)
2 - clear the CIDR
_Expected:_ 
ICMP type and code are disabled
_Actual:_
ICMP type and code are enabled

Issue #997, opened 2018-02-19
Ability to add a Firewall rule when incorrect IP 


chewlite:
Steps:
1. Go to Firewall
2. Go to Rules editing
3. Fill all mandatories correctly and

**_3.1_**  in "CIDR v4/v6" input IPv4/IPv6 without subnet mask length 
Expected result: IP is invalid, add-button is disabled
Actual result: IP is valid, add-button is enabled
![valid cidr](https://user-images.githubusercontent.com/32011943/36404533-ced7091a-161d-11e8-87cb-b306f3a27b03.png)

**_3.2_**  in "CIDR v4/v6" input invalid IPv4/IPv6 (with special symbols for example)
Expected result: IP is invalid, add-button is disabled
Actual result: IP is invalid, but add-button is enabled
![invalid cidr but add is enable](https://user-images.githubusercontent.com/32011943/36372061-74223486-1596-11e8-82a6-66895844fa86.png)

**Connected Feature ID:**
_firewall_rules_edit
vm_firewall_rules_edit_

ershovaelena:
TBD in #1019

tamazlykar:
Test on: 
tamazlykar/997-cidr

Connected Feature ID:
firewall_rules_edit
vm_firewall_rules_edit

chewlite:
@tamazlykar 
Tested on:
tamazlykar/997-cidr

tamazlykar:
The new library used in #1078 broke IE.

tamazlykar:
Please test same functionality in different browsers

Test on:
tamazlykar/997-cidr

chewlite:
@tamazlykar 
Tested on: tamazlykar/997-cidr

PR #1064, opened 2018-04-27
fix(sg): update and fix bugs in rule addition form

### LINKED ISSUE #1019: Ability to add a Firewall rule and no validation when incorrect Start and End ports
Steps:

1) Go to firewall rules editing
2) Fill Start port by greater value then End port, click add

Expected result: Start and End ports are invalid, add button is disabled
Actual result: Start and End ports are valid, add button is enabled

**Connected Feature ID:**
firewall_rules_edit
vm_firewall_rules_edit

tamazlykar:
This is a large set of bug fixes in the form for adding a firewall rule. For simplicity of maintenance, it was reorganized into a separate component. Now form uses Reactive Forms because this form requires a lot of logic that will be hard to maintain in Template-driven Form.

This fixes #1019, #1000, and 3.2 item at #997 

Additionally,
* added "Start Port" and "End Port"  floating labels for the corresponding fields
* added cleansing of the "Start Port" and "End Port" fields after creating the rule
* added validation of existing ICMP types and codes (there was a bug that allows us to create a rule with nonexistent type and code)
* changed the way of adding ICMP rule for better autocomplete support and predictable user experience. Now fields depend on each other. Initially, ICMP type and code fields disabled. After user entered correct CIDR, ICMP type becoming enabled and after correct entered ICMP type, ICMP code field becoming enabled

Added feature
* When filling the start/end port number, the value is duplicated to the end/start port if it was not touched by the user
* Start and end port has 4 different error messages:
  * if a value greater than maximum available
  * if a value less than minimum available
  * if a user set and then removed the value, so the field is empty
  * if start port greater than end port and vice versa

See live changes on:
tamazlykar/1019-firewall-rules


src/app/security-group/components/sg-rule-addition-form/sg-rule-addition-form.component.scss
```diff
@@ -0,0 +1,14 @@
+mat-form-field {
+  width: 150px;
+}
+
+form{
```
zolotyx: fix code style

src/app/security-group/components/sg-rule-addition-form/sg-rule-addition-form.component.ts
```diff
@@ -0,0 +1,414 @@
...
+  GetICMPV6TypeTranslationToken,
+  IcmpType,
+  icmpV4Types,
+  icmpV6Types
+} from '../../../shared/icmp/icmp-types';
+import 'rxjs/add/operator/startWith';
+import { FirewallRule } from '../../shared/models';
+
+
+function isExist(value: any): boolean {
```
zolotyx: get rid of this function. is confuses

src/app/security-group/components/sg-rule-addition-form/sg-rule-addition-form.component.ts
```diff
@@ -0,0 +1,414 @@
...
+  return !!value
+}
+
+function getIpVersionByCidr(cidr: string): IPVersion {
```
zolotyx: why can't you use Utils.cidrType directly?

src/app/security-group/components/sg-rule-addition-form/sg-rule-addition-form.component.html
```diff
@@ -0,0 +1,116 @@
...
+    <mat-form-field>
+      <input
+        matInput
+        type="number"
+        formControlName="startPort"
+        [errorStateMatcher]="portMatcher"
+        [placeholder]="'SECURITY_GROUP_PAGE.RULES.START_PORT' | translate"
+        required
+      >
+      <mat-error>
```
zolotyx: Can you please list out all the error messages here and show them using ngIf?
similar to other fields

src/app/security-group/components/sg-rule-addition-form/sg-rule-addition-form.component.ts
```diff
@@ -0,0 +1,414 @@
...
+  }
+
+  private getFormData(): FirewallRule {
+    const formModel = this.ruleForm.value;
+    const rule: FirewallRule = {
+      type: formModel.type,
+      protocol: formModel.protocol,
+      cidr: formModel.cidr
+    };
+    if (rule.protocol === NetworkProtocol.ICMP) {
```
zolotyx: Think about getting rid of mutation

src/app/security-group/components/sg-rule-addition-form/sg-rule-addition-form.component.ts
```diff
@@ -0,0 +1,414 @@
...
+      Validators.min(this.minPortNumber),
+      Validators.max(this.maxPortNumber)
+    ];
+    const startPort = this.portsForm.get('startPort');
+    const endPort = this.portsForm.get('endPort');
+    startPort.setValidators([...portCommonValidators, startPortValidator(endPort)]);
+    endPort.setValidators([...portCommonValidators, endPortValidator(startPort)]);
+  }
+
+  private onProtocolChange() {
```
zolotyx: can we add a handler for an event emitter of the protocol field?
same for other subscriptions

src/app/security-group/components/sg-rule-addition-form/sg-rule-addition-form.component.ts
```diff
@@ -0,0 +1,414 @@
...
+        this.updateDisabledStatusOnIcmpCodeField();
+      });
+  }
+
+  private updateFilteredIcmpTypes(value: number | string) {
+    this.filteredIcmpTypes = this.filterIcmpTypes(value);
+  }
+
+  private filterIcmpTypes(val: number | string): IcmpType[] {
+    if (!!val === false) {
```
zolotyx: replace with if(!val)

src/app/security-group/components/sg-rule-addition-form/sg-rule-addition-form.component.ts
```diff
@@ -0,0 +1,414 @@
...
+      return this.getIcmpTypes();
+    }
+    return this.getIcmpTypes().filter(el => {
+      const filterValue = val.toString().toLowerCase();
```
zolotyx: move this operation out fo the loop

src/app/security-group/components/sg-rule-addition-form/sg-rule-addition-form.component.ts
```diff
@@ -0,0 +1,414 @@
...
+        this.updateFilteredIcmpCodes(value);
+      });
+  }
+
+  private updateFilteredIcmpCodes(value: number | string) {
+    this.filteredIcmpCodes = this.filterIcmpCodes(value);
+  }
+
+  private filterIcmpCodes(val: number | string) {
+    if (!!val === false) {
```
zolotyx: replace with !val

src/app/security-group/components/sg-rule-addition-form/sg-rule-addition-form.component.ts
```diff
@@ -0,0 +1,414 @@
...
+      return this.getIcmpCodes();
+    }
+    return this.getIcmpCodes().filter(code => {
+      const filterValue = val.toString().toLowerCase();
```
zolotyx: optimize

src/app/security-group/components/sg-rule-addition-form/sg-rule-addition-form.component.ts
```diff
@@ -0,0 +1,414 @@
...
+    }
+    return cidrIpVersion === IPVersion.ipv6 ? icmpV6Types : icmpV4Types;
+  }
+
+  private getIcmpCodes(): number[] {
+    let type = this.icmpForm.get('icmpType').value;
+    if (!type) {
+      return [];
+    }
+    type = +type;
```
zolotyx: get rid of mutation

src/app/security-group/shared/validators/cidr.validator.ts
```diff
@@ -0,0 +1,9 @@
+import { AbstractControl, ValidatorFn } from '@angular/forms';
+import { Utils } from '../../../shared/services/utils/utils.service';
+
+export function cidrValidator(): ValidatorFn {
+  return (control: AbstractControl): { [key: string]: any } => {
```
zolotyx: I believe that we can use some proper typing { [key: string]: any }

zolotyx: Same for other validators

src/app/shared/services/utils/utils.service.ts
```diff
@@ -89,12 +89,18 @@ export class Utils {
   };
 
   public static cidrIsValid(range: string): boolean {
+    if (range === null) {
```
zolotyx: add null to range type

src/app/security-group/components/sg-rule-addition-form/sg-rule-addition-form.component.scss
tamazlykar: +

src/app/security-group/components/sg-rule-addition-form/sg-rule-addition-form.component.ts
tamazlykar: ok

src/app/security-group/components/sg-rule-addition-form/sg-rule-addition-form.component.ts
tamazlykar: I was thinking that will be more clear naming for function. But okay I'll change it

src/app/security-group/components/sg-rule-addition-form/sg-rule-addition-form.component.ts
tamazlykar: ok

src/app/security-group/components/sg-rule-addition-form/sg-rule-addition-form.component.ts
tamazlykar: ok

src/app/security-group/components/sg-rule-addition-form/sg-rule-addition-form.component.ts
tamazlykar: ok

src/app/security-group/components/sg-rule-addition-form/sg-rule-addition-form.component.ts
tamazlykar: done

src/app/security-group/shared/validators/cidr.validator.ts
tamazlykar: changed to ValidationErrors | null

src/app/shared/services/utils/utils.service.ts
tamazlykar: ok

src/app/security-group/sg-rules/sg-rules.component.html
```diff
@@ -6,254 +6,87 @@
...
+      </mat-form-field>
+      <mat-form-field class="form-select-control">
+        <mat-select
+          multiple="true"
+          [placeholder]="'SECURITY_GROUP_PAGE.RULES.SELECT_TYPE' | translate"
+          name="selectedTypes"
+          [(ngModel)]="selectedTypes"
+          (selectionChange)="filter()"
+        >
+          <mat-option *ngFor="let z of types" [value]="z.value">
```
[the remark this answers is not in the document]
tamazlykar: ok

src/app/security-group/components/sg-rule-addition-form/sg-rule-addition-form.component.ts
tamazlykar: ok

src/app/security-group/components/sg-rule-addition-form/sg-rule-addition-form.component.ts
tamazlykar: Yes, we can, but the idea of reactive forms to put eject logic from a template and place it to a component. Also using this way I receive an observable and I can transform it before use. It will be much clearer to use this approach.

src/app/security-group/components/sg-rule-addition-form/sg-rule-addition-form.component.ts
tamazlykar: ok

src/app/security-group/components/sg-rule-addition-form/sg-rule-addition-form.component.html
tamazlykar: It has already been discussed and decided to leave it as it is

tamazlykar:
@zolotyx @vanekbr Requested changes provided, some of them not fulfilled see comments. Please review changes.
There also some improvements in ICMP code and type validators.

PR #1077, opened 2018-05-14
940 input type number TEMP

pm-Egor-Genning:


src/ngrx.entity.dictionary.ts
```diff
@@ -0,0 +1,5 @@
+import { DictionaryNum } from '@ngrx/entity/src/models';
+
+export declare abstract class Dictionary<T> implements DictionaryNum<T> {
```
zolotyx: why not to use https://github.com/ngrx/platform/blob/master/modules/entity/src/models.ts#L25?

src/app/shared/components/input-type-number/input-type-number.component.ts
```diff
@@ -0,0 +1,162 @@
...
+} from '@angular/core';
+import { FormControl, ValidatorFn, Validators } from '@angular/forms';
+import { Subscription } from 'rxjs/Subscription';
+
+@Component({
+  selector: 'cs-input-type-number',
+  templateUrl: 'input-type-number.component.html',
+  styleUrls: ['input-type-number.component.scss']
+})
+export class InputTypeNumberComponent implements OnChanges, OnDestroy, AfterViewInit {
```
zolotyx: why not to use https://angular.io/api/forms/ControlValueAccessor?

src/app/shared/components/input-type-number/input-type-number.component.html
```diff
@@ -0,0 +1,32 @@
+<mat-form-field [ngClass]="className">
+  <input
```
zolotyx: can you implement just an input without mat-form-field and mat-error?

src/ngrx.entity.dictionary.ts
tamazlykar: This is not in public api from this library. It produces errors.

zolotyx:
Closing this PR, because @tamazlykar is implementing another solution

PR #1078, opened 2018-05-15
fix(firewall): replace library for cidr manipulation

### LINKED ISSUE #997: Ability to add a Firewall rule when incorrect IP

tamazlykar:
Fixes #997 
I changed 'ip-address' library to 'cidr-regex'. That suits our needs most. And wrote some test.


src/app/shared/services/utils/utils.service.spec.ts
```diff
@@ -131,4 +132,89 @@ describe('Utils service', () => {
...
+    for (const value of validValues) {
+      expect(Utils.cidrIsValid(value)).toBe(true);
+    }
+    for (const value of invalidValues) {
+      expect(Utils.cidrIsValid(value)).toBe(false);
+    }
+  });
+
+  it('should return proper CIDR type', () => {
+    const validV4Values = [
```
vanekbr: Move addresses const declarations out of test to the root of the file to avoid code duplication

src/app/shared/services/utils/utils.service.ts
```diff
@@ -88,22 +88,21 @@ export class Utils {
...
+  public static cidrType(cidr: string | null): IPVersion | null {
+    if (!Utils.cidrIsValid(cidr)) {
       return null;
     }
-    const ipAddressType = range.match(':') ? ipaddr.Address6 : ipaddr.Address4;
-    const cidr = new ipAddressType(range);
-    return cidr.isValid() && (cidr.v4 ? IPVersion.ipv4 : IPVersion.ipv6);
+    const isIPv4 = cidrRegex.v4({exact: true}).test(cidr);
+
+    if (isIPv4) {
```
vanekbr: Use ternary operator here
`return isIPv4 ? IPVersion.ipv4 : IPVersion.ipv6`

PR #1079, opened 2018-05-15
fix: add query parameter for redirect to shared SG

### LINKED ISSUE #1016: Grouping isn't saved, Delete buttons are absent, redirect to wrong tab when shared SG rules edit through VM sidebar
Steps

1) Log in and go to sidebar of VM with shared security group
2) Open the rule view and select some parameters of grouping
3) Go to rules edit

Expected result: redirect to Shared SGs tab, there are delete buttons, grouping is saved
Actual result: redirect to Firewall Templates tab, there are no delete buttons, grouping isn't saved
(If Shared SGs tab was opened before going to VMs then redirect will be to this tab)

rules view vm
rules edit vm

Connected Feature ID:
vm_firewall_rules_edit

tamazlykar:
Fixes #1016 
But there is another redirect performs after navigating and it takes some time. An issue for this will be created and fixed.

### DIFF
--- a/src/app/security-group/sg-rules/sg-rules.component.ts
+++ b/src/app/security-group/sg-rules/sg-rules.component.ts
@@ -179,7 +179,7 @@ export class SgRulesComponent implements OnInit, OnChanges {
               ], {
                 queryParams: {
                   vm: this.vmId,
-                  viewMode: 'shared'
+                  viewMode: SecurityGroupType.Shared
                 }
               });
               this.onCloseDialog.emit();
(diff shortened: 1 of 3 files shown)


src/app/security-group/sg-rules/sg-rules.component.ts
```diff
@@ -174,7 +177,10 @@ export class SgRulesComponent implements OnChanges {
               this.router.navigate([
                 'security-group', this.securityGroup.id, 'rules'
               ], {
-                queryParams: { vm: this.vmId }
+                queryParams: {
+                  vm: this.vmId,
+                  viewMode: 'shared'
```
vanekbr: Use enum value here: `viewMode: SecurityGroupViewMode.Shared`

tamazlykar:
@vanekbr fixed :)

PR #1082, opened 2018-05-18
 feat: cross-browser validation of values in a numeric input field

### LINKED ISSUE #940: Buttons for change disk size are not displayed when create or resize disk - MS Edge, IE
Browser: Edge (ver. 41.16299.15.0), IE (ver. 11.192.16299.0)

Steps:
1. Log in and go to Storage / VM sidebar.
2. Create new storage or edit old.
3. Select editable disk offering.

Actual result: Buttons of change disk size is not displayed;
Expected result: Buttons of change disk size is displayed; (Including ROOT disks)
Connected Feature ID: vm_disk_resize, storage_disk_resize, storage_disk_create

Screenshots:

tamazlykar:
Fixes #940 
The implemented directive used to prevent user to the entry of the text values into the numeric input field and to limit the possibility of entering numbers beyond the boundaries settled by "csMaxValue" and "csMinValue".
Problem was that it wasn't validated by some browsers like IE.

Also, there is a new requirement for number inputs that have some valid ranges.
* input can't be empty, min value should be placed instead (exception: new untouched form)
* input doesn't allow you enter lesser value than a minimum or greater than a maximum


package.json
```diff
@@ -50,6 +50,7 @@
     "rxjs": "^5.5.2",
     "showdown": "^1.8.4",
     "sprintf-js": "^1.1.1",
+    "typescript-debounce-decorator": "^0.0.17",
```
zolotyx: use lodash instead

src/app/shared/directives/input-type-number.directive.ts
```diff
@@ -0,0 +1,135 @@
...
+  provide: NG_VALUE_ACCESSOR,
+  useExisting: forwardRef(() => InputTypeNumberDirective),
+  multi: true
+};
+
+@Directive({
+  selector: 'input[type=number]',
+  providers: [INPUT_NUMBER_ACCESSOR]
+})
+export class InputTypeNumberDirective implements ControlValueAccessor {
```
zolotyx: add unit tests

PR #1084, opened 2018-05-21
fix: get rid of cidr library

### LINKED ISSUE #997: Ability to add a Firewall rule when incorrect IP

tamazlykar:
Fixes #997 
The library that was used broke IE, so we got rid of it.
Now CIDR validation doesn't depend on third-party libraries.

### DIFF
--- a/src/app/security-group/components/sg-rule-addition-form/sg-rule-addition-form.component.ts
+++ b/src/app/security-group/components/sg-rule-addition-form/sg-rule-addition-form.component.ts
@@ -9,7 +9,6 @@ import {
   Validators
 } from '@angular/forms';
 import { Subscription } from 'rxjs/Subscription';
-import 'rxjs/add/operator/startWith';
 import { TranslateService } from '@ngx-translate/core';
 import { ErrorStateMatcher } from '@angular/material';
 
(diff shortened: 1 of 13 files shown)


src/app/security-group/components/sg-rule-addition-form/sg-rule-addition-form.component.ts
```diff
@@ -9,12 +9,12 @@ import {
   Validators
 } from '@angular/forms';
 import { Subscription } from 'rxjs/Subscription';
+import 'rxjs/add/operator/startWith';
```
zolotyx: Let's add this to https://github.com/bwsw/cloudstack-ui/blob/master/src/polyfills.ts

tamazlykar: This is a redundant import, I've deleted it

PR #1107, opened 2018-06-14
fix(vm): update handling an affinity group creation response

### LINKED ISSUE #1095: New affinity group does not connect to the virtual machine
### Steps to reproduce
1. Open Vm sidebar
2. Select edit affinity group
3. Choose "Create a new group"
4. Enter new name and press assign

### Actual result

When you open VM sidebar and want to attach new affinity group to VM the affinity group created but not attached.

### Expected result

New affinity group created and attached to VM.

### Environment
<pre><code>
Browser:
- [ ] Chrome version XX
- [ ] Firefox version XX
- [ ] Safari version XX
- [ ] IE version XX
- [ ] Edge version XX

Others:

</code></pre>

### Connected feature ID

vm_afgroup_change

tamazlykar:
Fixes #1095

### DIFF
--- a/src/app/shared/services/affinity-group.service.ts
+++ b/src/app/shared/services/affinity-group.service.ts
@@ -4,7 +4,7 @@ import { Observable } from 'rxjs/Observable';
 import { VirtualMachine } from '../../vm/shared/vm.model';
 import { BackendResource } from '../decorators';
 
-import { AffinityGroup } from '../models';
+import { AffinityGroup, AsyncJob } from '../models';
 import { AffinityGroupType } from '../models/affinity-group.model';
 import { AsyncJobService } from './async-job.service';
 import { BaseBackendCachedService } from './base-backend-cached.service';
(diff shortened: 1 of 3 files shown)


src/app/shared/services/affinity-group.service.ts
```diff
@@ -31,7 +31,8 @@ export class AffinityGroupService extends BaseBackendCachedService<AffinityGroup
 
   public create(params: AffinityGroupCreationData): Observable<AffinityGroup> {
     return super.create(params)
-      .switchMap(job => this.asyncJob.queryJob(job.jobid, this.entity, this.entityModel));
+      .switchMap(job => this.asyncJob.queryJob(job.jobid, this.entity, this.entityModel))
+      .map((job: AsyncJob<any>) => job.jobresult.affinitygroup);
```
zolotyx: is it possible to get rid of **any**

tamazlykar: Due to the fact that we began to use interfaces when switching to ngrx, the services began to work a little differently. In this case, AsyncJob generic class will not work correctly.

PR #1117, opened 2018-06-15
fix(sg): do not use cache for security groups

### LINKED ISSUE #1098: Can not delete a private security group without refreshing the page
### Steps to reproduce
1. Create VM
2. Delete this VM (don't expunge)
3. Wait until deleted VM disappears approx. 1-2 min (you need to refresh the data pressing on VM link in the menu)
4. Go to Firewall and Private SG tab
5. Try to delete the SG belonging to removed VM

### Actual result

Unable to delete Private SG that belongs to removed VM without refreshing the page. There is no such action in menu.

### Expected result

The ability to remove Private SG, which belongs to a removed VM, without updating the page. 

### Environment
<pre><code>
Browser:
- [ ] Chrome version XX
- [ ] Firefox version XX
- [ ] Safari version XX
- [ ] IE version XX
- [ ] Edge version XX

Others:

</code></pre>

tamazlykar:
Fixes #1098 
What is going on:
After the virtual machine is deleted, it is still in the list with the status removed. At this time, the security group still believes that she has a VM. After a while, CloudStack will remove the virtual machine from the list and only after this removal the security group will not have attached virtual machines and we can delete it.

### DIFF
--- a/src/app/security-group/sg-actions/sg-actions-component/sg-actions.component.ts
+++ b/src/app/security-group/sg-actions/sg-actions-component/sg-actions.component.ts
@@ -25,13 +25,13 @@ export class SecurityGroupActionsComponent implements OnInit {
 
 
   public ngOnInit() {
+    this.actions = this.securityGroupActionService.actions;
+
     if (this.securityGroup && this.securityGroup.preselected) {
-      this.actions = this.securityGroupActionService.actions
-        .filter(action => action.command !== SecurityGroupActionType.Delete);
-    } else {
-      this.actions = this.securityGroupActionService.actions;
+      this.actions = this.actions.filter(action => action.command !== SecurityGroupActionType.Delete);
     }
   }
+
 
   public onAction(action): void {
     switch (action.command) {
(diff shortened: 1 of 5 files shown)


tamazlykar:
PR updated. I remove the delete action for a system templates. These templates define by an administrator and can not be removed in any case.

src/app/security-group/sg-actions/sg-actions-component/sg-actions.component.ts
```diff
@@ -1,28 +1,38 @@
...
   ) {
   }
 
+
+  public ngOnInit() {
+    if (this.securityGroup && this.securityGroup.preselected) {
+      this.actions = this.securityGroupActionService.actions
+        .filter(action => action.command !== SecurityGroupActionType.Delete);
+    } else {
+      this.actions = this.securityGroupActionService.actions;
```
vanekbr: Move this line to the top of the ngOnInit method and then filter `this.action` by the condition

PR #1126, opened 2018-06-22
fix(accounts): hide menu instead of remove to save block height

### LINKED ISSUE #1115: The current user's account card is smaller than the rest
### Steps to reproduce
1. Go to Accounts section

### Actual result

The current user's account card is smaller than the rest

### Expected result

All cards have the same size

### Environment
<pre><code>
Browser:
- [ ] Chrome version XX
- [ ] Firefox version XX
- [ ] Safari version XX
- [ ] IE version XX
- [ ] Edge version XX

Others:

</code></pre>

### Screenshots

screenshot from 2018-06-15 15-11-39

tamazlykar:
Fixes #1115 
Used visibility instead of ngif for account card's menu button so that the element was not removed from the DOM and passed its size.

### DIFF
--- a/src/app/template/template-sidebar/template-os-icon/template-os-icon.component.html
+++ b/src/app/template/template-sidebar/template-os-icon/template-os-icon.component.html
@@ -5,25 +5,13 @@
   mat-card-avatar
 >
   <ng-template [ngSwitchCase]="'Linux'">
-    <img
-      src="img/os/linux.png"
-      width="36"
-      height="36"
-    >
+    <img src="img/os/linux.png">
   </ng-template>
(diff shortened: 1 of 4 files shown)


src/app/template/template-sidebar/template-os-icon/template-os-icon.component.html
```diff
@@ -5,25 +5,13 @@
   mat-card-avatar
 >
   <ng-template [ngSwitchCase]="'Linux'">
-    <img
-      src="img/os/linux.png"
-      width="36"
-      height="36"
-    >
+    <img src="img/os/linux.png">
```
vanekbr: Create css class for these images, do not style the `img` tag

PR #1154, opened 2018-07-13
fix(vm): fetch updated VM after adding or removing secondary IP

### LINKED ISSUE #1134: Error message when adding an additional IP
### Steps to reproduce
1. Select a VM
2. Go to the Network tab
3. Add a secondary IP

### Actual result

An error appears, the rule is added, but it is visible after the page is updated.

### Expected result

No error appears and a rule is visible right after creation.

### Environment
<pre><code>
Browser:
- [ ] Chrome version XX
- [ ] Firefox version XX
- [ ] Safari version XX
- [ ] Opera version XX
- [ ] IE version XX
- [ ] Edge version XX

Others:

</code></pre>

### Connected feature ID

vm_NIC_secondIP_add

### Screenshots

screenshot from 2018-06-27 11-21-07

tamazlykar:
Fixes #1134

### DIFF
--- a/src/app/vm/shared/vm.service.ts
+++ b/src/app/vm/shared/vm.service.ts
@@ -2,24 +2,15 @@ import { HttpClient } from '@angular/common/http';
 import { Injectable } from '@angular/core';
 import { Observable } from 'rxjs/Observable';
 import { BackendResource } from '../../shared/decorators';
-import {
-  OsType,
-  ServiceOffering,
-  Volume
-} from '../../shared/models';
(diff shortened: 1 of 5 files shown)


src/app/vm/shared/vm.service.ts
```diff
@@ -136,7 +127,8 @@ export class VmService extends BaseBackendService<VirtualMachine> {
 
   public addIpToNic(nicId: string): Observable<any> {
```
zolotyx: Please replace any with a proper type if possible

PR #1179, opened 2018-07-30
refactor(account-model): get rid of BaseModel for Account User

HeyRoach:


tamazlykar:
run tests

HeyRoach:
All fields optional after this commit
https://github.com/bwsw/cloudstack-ui/commit/55a6ae25a2ba9698e2c3dcea9516343d2417b52d

tamazlykar:
I don't understand why she did that.
@zolotyx what do you think about this?

zolotyx:
I think that the purpose of removing BaseModel classes is to move to interfaces.
@HeyRoach , please change that to an interface and check if we do need to have al the fields optional

src/app/account/account-sidebar/account-users/account-users.component.ts
```diff
@@ -49,7 +49,7 @@ export class AccountUsersComponent {
     this.dialog.open(AccountUserPasswordFormContainerComponent, {
       width: '375px',
       data: {
-        userId: user.id
+        user: user
```
tamazlykar: You don't need to duplicate user twice eg. `data: { user }`

src/app/account/account-sidebar/account-users/account-user-edit.component.ts
```diff
@@ -44,7 +44,7 @@ export class AccountUserEditComponent implements OnInit {
   }
 
   public prepareData(data: AccountUserForm): AccountUser {
-    const result: AccountUser = new AccountUser();
+    const result: AccountUser = {} as AccountUser;
```
zolotyx: remove **as**

src/app/account/account-sidebar/account-users/account-user-edit.component.ts
```diff
@@ -55,6 +55,6 @@ export class AccountUserEditComponent implements OnInit {
     if (data.timezone) {
       result.timezone = data.timezone.geo;
     }
-    return result;
+    return result as AccountUser;
```
zolotyx: ditto

PR #1194, opened 2018-08-10
feat: keep sidenav visibility state on the server for every user

### LINKED ISSUE #944: Keep the position of a left-side bar (collapsed/expanded)
**Acceptance criteria:**

The position of a left-side bar (collapsed/expanded) should be remembered after 
- reloading the page
- going to another section 
- within the section
- login/logout

**Connected feature ID:** _sidenav_show_

tamazlykar:
Closes #944 

We omit the result of setting the tag value representing sidenav status on the server.
This is required so that the UI reacts instantly and does not wait until an answer comes from the server.
Downsides: if the tag is not set, the user selected state will not be saved (this can happen very rarely)


src/app/core/components/sidenav/sidenav.component.ts
```diff
@@ -83,8 +75,8 @@ export class SidenavComponent extends WithUnsubscribe() implements AfterViewInit
     return new Date().getFullYear().toString();
   }
 
-  public toggleDrawer(): void {
-    this.layoutService.toggleDrawer();
+  public closeSidenav(): void {
+    this.store.dispatch(new layoutActions.CloseSidenav);
```
zolotyx: please, add brackets '()'

src/app/home/home.component.ts
```diff
@@ -16,10 +15,10 @@ import * as serviceOfferingActions from '../reducers/service-offerings/redux/ser
 })
 export class HomeComponent extends WithUnsubscribe() implements OnInit {
   public disableSecurityGroups = false;
+  public showSidenav$ = this.store.select(layoutSelectors.getIsShowSidenav);
```
zolotyx: why do you use this language?
"is show sidenav" why not to use" "is sidenav visible"

tamazlykar:
Changes added

PR #1215, opened 2018-08-17
fix(tags): add actions for tag consisting of spaces

HeyRoach:


zolotyx:
Maybe just trim the value before saving it?

src/app/shared/directives/forbidden-values-validator.ts
```diff
@@ -5,7 +5,7 @@ export function forbiddenValuesValidator(
   forbiddenValues: Array<string>
 ): ValidatorFn {
   return (c: AbstractControl) => {
-    const isValid = !forbiddenValues.find(_ => _ === c.value);
+    const isValid = c.value && !forbiddenValues.find(_ => _ === c.value.trim());
```
tamazlykar: screenshot from 2018-08-23 10-10-57

src/i18n/en.json
```diff
@@ -1565,7 +1565,8 @@
     "TAG_KEY": "Key",
     "TAG_VALUE": "Value",
     "TAG_ALREADY_EXISTS": "Tag already exists",
-    "EDIT_TAG": "Edit tag"
+    "EDIT_TAG": "Edit tag",
+    "TAG_INVALID": "Value shouldn't started from space"
```
tamazlykar: Use name that describes a translation string

tamazlykar:
@ershovaelena 
screenshot from 2018-08-23 10-15-02
Tags have different number of spaces but looks identical

bwsw:
@HeyRoach Trimming is OK, by the way. No problem with that. @zolotyx thnx. 
@ershovaelena why you decided to go with error rather than just trimming?

bwsw:
@HeyRoach don't change the behavior now. Just show an error if first char is a space char.

src/i18n/en.json
```diff
@@ -1565,7 +1565,8 @@
     "TAG_KEY": "Key",
     "TAG_VALUE": "Value",
     "TAG_ALREADY_EXISTS": "Tag already exists",
-    "EDIT_TAG": "Edit tag"
+    "EDIT_TAG": "Edit tag",
+    "TAG_START_FROM_SPACE": "Value shouldn't started from space"
```
bwsw: @HeyRoach 
The tag keys and values shouldn't start with "space" symbol.

PR #1216, opened 2018-08-17
fix(account): show copy icon in account's sidebar in IE

HeyRoach:


src/app/shared/components/parameters-pair/parameters-pair.component.scss
```diff
@@ -22,6 +22,7 @@
   cs-input-group {
     mat-form-field {
       padding-top: 0.2em;
+      max-width: 100px;
```
tamazlykar: can there be another solution that does not the component be constrained?

PR #1221, opened 2018-08-23
 fix: improve configureSidenav configuration

### LINKED ISSUE #1166: ConfigureSidebar feature does not work as expected
### Expected behavior

The configureSidebar settings should configure a sidebar order.  If we remove an element from the array, then this menu section must be disabled in the application menu setup mode.

### Actual behavior

Menu items don't follow the order in the configureSidebar setting.
If we remove an item from the configureSidebar Array them this section disappears completely even from menu setup mode, but you can go there by URL.

### Environment
<pre><code>
Browser:
- [ ] Chrome version XX
- [ ] Firefox version XX
- [ ] Safari version XX
- [ ] Opera version XX
- [ ] IE version XX
- [ ] Edge version XX

Others:

</code></pre>

### Connected feature ID

sidebar_reorder
config_configure_sidebar

tamazlykar:
Fixes #1166


src/app/core/components/sidenav/sidenav.component.ts
```diff
@@ -26,6 +27,7 @@ export class SidenavComponent extends WithUnsubscribe() implements OnInit, OnDes
 
   public routes: Array<SidenavRoute> = cloneDeep(sidenavRoutes);
   public nonDraggableRoutes = nonDraggableRoutes;
+  public alwaysVisibleRouteId: SidenavRouteId = 'VMS';
```
zolotyx: I think that It's better to create a method for that. Something like canBeHidden or isAlwaysVisible

PR #1306, opened 2018-10-03
fix(grouping): update draggable select to the latest material

wowshakhov:


src/app/shared/components/draggable-select/draggable-select.component.html
```diff
@@ -38,7 +38,7 @@
     class="mat-select-panel draggable-select-panel {{ _getPanelTheme() }}"
     [ngClass]="panelClass"
     [@transformPanel]="multiple ? 'showing-multiple' : 'showing'"
-    (@transformPanel.done)="_onPanelDone()"
+    (@transformPanel.done)="_panelDoneAnimatingStream.next($event.toState)"
```
wowshakhov: As per https://github.com/angular/material2/blob/18d0fa869ebe4da43461479e7ac67607260c1d58/src/lib/select/select.html#L38

PR #1317, opened 2018-10-09
fix(vm-default-name): load default name

wowshakhov:


src/app/root-store/server-data/user-tags/user-tags.reducer.ts
```diff
@@ -83,3 +84,10 @@ export function reducer(state = initialState, action: UserTagsActionsUnion): Use
     }
   }
 }
+
+export const getUserTagsState = createFeatureSelector<UserTagsState>('userTags');
```
wowshakhov: Moved getState and isLoading to `.reducer.ts` for consistency, because everywhere else these are placed in `.reducer.ts`.

src/app/vm/vm-creation/containers/vm-creation.container.ts
```diff
@@ -197,7 +198,6 @@ export class VmCreationContainerComponent implements OnInit {
   private getDefaultVmName(): Observable<string> {
     return this.store.pipe(
       select(UserTagsSelectors.getLastVMId),
```
wowshakhov: If you use `first()`, lastVMId from getUserTags response is ignored.

src/app/vm/vm-creation/containers/vm-creation.container.ts
```diff
@@ -37,6 +37,7 @@ import * as vmActions from '../../../reducers/vm/redux/vm.actions';
 import * as fromVMs from '../../../reducers/vm/redux/vm.reducers';
 import * as zoneActions from '../../../reducers/zones/redux/zones.actions';
 import * as fromZones from '../../../reducers/zones/redux/zones.reducers';
+import * as fromUserTags from '../../../root-store/server-data/user-tags/user-tags.selectors';
```
tamazlykar: import { UserTagsSelectors } from '../../../root-store';

src/app/vm/vm-creation/security-group/vm-creation-security-group-data.ts
```diff
@@ -24,6 +20,6 @@ export class VmCreationSecurityGroupData {
   private constructor(mode?: VmCreationSecurityGroupMode, rules?: Rules, securityGroups?: Array<SecurityGroup>) {
     this.mode = mode;
     this.rules = rules;
-    this.securityGroups = securityGroups;
+    this.securityGroups = securityGroups || [];
```
tamazlykar: Use default parameter value instead.

Issue #1235, opened 2018-08-28
Rework the left-side menu


ershovaelena:
This issue covers the following requirements:
[REQ_GEN_006] A user should be able to see the left-side menu and navigate through sections.

**Acceptance criteria:**

**1. Change the structure of the left-side menu.** 
1) There should be two 1-level sections (Virtual Machines, Accounts).
2) When a user clicks on 1-level section, a detailed menu opens (2-level sections).
3) User can come back to 1-level section by clicking on sandwich button
3) The structure of the menu:

Virtual Machines
- Virtual machines
- Storage
- Images
- Snapshots
- Firewall
- SSH keys
- Logout

Accounts
- Accounts
- Activity Log
- Settings
- Logout
4) Use as an example (see the gif attached). The only difference is that we do not have a fixed up-side panel, so sandwich button that opens and hides 1-level menu should be presented in both modes (see a gif).
5) VM (2 level) should be opened by default
6) Login information should be displayed in 2 level
7) Logout should be displayed in both levels (1st and 2nd) 

**2. Remove features (as this structure changes the way the menu works) from the code + from config + from documentation**
1) Hide/collapse sidebar (sidenav_show)
2) Reorder sidebar (sidebar_reorder)

**3. Please, approve mockups with PM before implementation**

**Connected feature ID:** sidenav, sidenav_show, sidebar_reorder

**Example**

![menu](https://user-images.githubusercontent.com/26703076/45133634-8e0f3a80-b1c0-11e8-8637-2847bd688384.gif)

chewlite:
@tamazlykar 
Tested on 1333

Issue #1061, opened 2018-04-18
Add Prettier for better formatting


tamazlykar:
This is a feature request to add Prettier in the project.
Currently, tslint and .editorconfig can do some code style normalize but they do not enough.

Using prettier could normalize code style more, avoid difference, it's much more useful to allow different IDE.

The plan
1. [x] Research whether style guide can be combined with Prettier
2. [x] Check out google and Airbnb style guides (optional)
3. [x] Discuss with a team which styleguide to use (if to use) and decide (optional)
4. [x] Configure tslint
5. [x] Configure codelyzer
6. [x] Configure Prettier
7. [x] Fix lint errors (if needed)
8. [ ] Create pre-commit hook or update CI.

ershovaelena:
@tamazlykar when this issue is closed, provide @Nadya913 @rennervo @chewlite  with the information about style guide.

tamazlykar:
During implementing this task I find a bug that in access VM dialog in the webshell tab, ssh key not appears. 
I'll fix it here.

tamazlykar:
The second bug about changing compute offering for VM.
Cannot change compute offering for VM even if I have a lot of available resources. The bug only about CPU number.

Prerequisites:
- An account has a limitation to CPU number equal 3.
- An account doesn't have any VM.
Steps:
1. Create a VM with 1 core so 2 more will be available for you after creation.
2. Try to change a compute offering for created VM. You should see an error that the service offering cannot be selected because it doesn't fit the account resources available.

tamazlykar:
The third bug about filtering VM by groups. This was not worked at all.

chewlite:
@tamazlykar 

All bug fixes from comments are tested in all browsers.
Surface testing of the entire system was performed in Chrome.

Fails:
1) The badge on the notification bell with the number of unwatched operations does not disappear after opening the bell

_Steps:_ 
1. Start/stop any VM
2. Wait until the operation will complete
3. Open the notification bell
_Actual:_ the badge isn't disappeared
_Expected:_ the badge is disappeared

2) The calendar isn't displayed in Snapshots and Activity log
![2018-10-19 13-08-19](https://user-images.githubusercontent.com/32011943/47200359-1415c800-d3a0-11e8-8a96-18458b069a5d.png)

chewlite:
@tamazlykar 

Fails from the comment above are tested

chewlite:
@tamazlykar

lack:2

[...]

PR #1333, opened 2018-10-15
feat(navigation): rework navigation menu

tamazlykar:
Implements #1235


src/app/core/nav-menu/components/app-nav/app-nav.component.html
```diff
@@ -0,0 +1,24 @@
+<div>
+  <cs-menu-header (buttonClicked)="close.emit(true)"></cs-menu-header>
```
wowshakhov: Why emit `true`?

src/app/core/nav-menu/nav-menu.service.ts
```diff
@@ -0,0 +1,58 @@
...
+
+  public getSubroutes(): Observable<Subroute[]> {
+    return this.getCurrentRoute().pipe(
+      map(route => route.subroutes)
+    );
+  }
+
+  public getCurrentRoute(): Observable<Route> {
+    return this.getCurrentSubroute().pipe(
+      filter(subroute => !!subroute),
```
wowshakhov: `filter(Boolean)`

src/app/core/nav-menu/nav-menu.service.ts
```diff
@@ -0,0 +1,58 @@
...
+    return of(appNavRoutes);
+  }
+
+  public getSubroutes(): Observable<Subroute[]> {
+    return this.getCurrentRoute().pipe(
+      map(route => route.subroutes)
+    );
+  }
+
+  public getCurrentRoute(): Observable<Route> {
```
wowshakhov: You can also test `getCurrentRoute`

src/app/core/nav-menu/nav-menu.service.spec.ts
```diff
@@ -0,0 +1,45 @@
...
+  it('should return a subroutes of the virtual-machines route based on URL', () => {
+    const url = '/templates/template/884adbea-ae67-4aba-86c0-c4f794c5343b/details?viewMode=Template';
+    store.setState(url);
+    const expectedRoute = appNavRoutes.find(route => route.id === 'virtual-machines');
+    service.getSubroutes().subscribe(subroutes => {
+      expect(subroutes).toBe(expectedRoute.subroutes);
+    })
+  });
+
+  it('should return a subroutes of the accounts route based on URL', () => {
```
wowshakhov: should return ~~a~~ subroutes

src/app/core/nav-menu/nav-menu.service.spec.ts
```diff
@@ -0,0 +1,45 @@
...
+    store = TestBed.get(Store);
+  });
+
+  it('should return an array of routes', () => {
+    service.getRoutes().subscribe(routes => {
+      expect(routes).toBe(appNavRoutes);
+    });
+  });
+
+  it('should return a subroutes of the virtual-machines route based on URL', () => {
```
wowshakhov: should return ~~a~~ subroutes (without "a")

src/app/core/nav-menu/components/license/license.component.ts
```diff
@@ -0,0 +1,34 @@
...
+    }
+
+    a {
+      color: inherit;
+    }
+  `]
+})
+
+export class LicenseComponent {
+  public getCurrentYear(): string {
```
MaverickW: Do you guys have some kind of style guide? Why are you using `public` modifier explicitly here, but for properties like `routes: Route[];` you omit them?

src/app/core/nav-menu/components/menu-header/menu-header.component.ts
```diff
@@ -0,0 +1,14 @@
...
+  selector: 'cs-menu-header',
+  changeDetection: ChangeDetectionStrategy.OnPush,
+  templateUrl: './menu-header.component.html',
+  styleUrls: ['./menu-header.component.scss']
+})
+export class MenuHeaderComponent {
+  @Input()
+  username: string;
+  @Output()
+  buttonClicked = new EventEmitter<boolean>();
```
MaverickW: Same thing as for `close = new EventEmitter<boolean>();` in `app-nav.component.ts`, do you really need `boolean` here?

src/app/core/core.module.ts
```diff
@@ -1,35 +1,35 @@
...
+  declarations: [
+    AppNavComponent,
+    SectionNavComponent,
+    LicenseComponent,
+    MenuHeaderComponent,
+  ],
+  providers: [
+    ConfigValidationService,
+    SnackBarService,
+    NavMenuService
```
MaverickW: Seems like `,` is missed here.

src/app/core/nav-menu/components/license/license.component.ts
```diff
@@ -0,0 +1,34 @@
+import { ChangeDetectionStrategy, Component } from '@angular/core';
+
+@Component({
+  selector: 'cs-license',
+  changeDetection: ChangeDetectionStrategy.OnPush,
+  template: `
```
MaverickW: Are there any rules regarding inline templates and styles? How do you decide to make this component with inline template and styles instead of putting them into separate files?

src/app/core/nav-menu/components/license/license.component.ts
```diff
@@ -0,0 +1,34 @@
...
+    <div class="license">
+      {{ 'NAVIGATION_SIDEBAR.LICENSE.LICENSE_1' | translate: { year: getCurrentYear() } }}
+      <a href="https://bitworks.software/" target="_blank">
+        {{ 'NAVIGATION_SIDEBAR.LICENSE.LICENSE_2' | translate }}
+      </a>
+      <br>
+      {{ 'NAVIGATION_SIDEBAR.LICENSE.LICENSE_3' | translate }}
+    </div>
```
MaverickW: It is not a good way to format content using `<br>` and separate translations for each line, because text can have different order and formatting for different languages.

For the lines, I suggest using wrappers like `<p>` or `<div>` instead of `<br>`.

As far as I know, you can put links and markup right inside of translation and pass all required data using translation params (year, link etc.). Isn't it?

src/app/core/nav-menu/components/app-nav/app-nav.component.scss
```diff
@@ -0,0 +1,28 @@
...
+    display: flex;
+    align-items: center;
+    color: inherit;
+    text-decoration: none;
+
+    mat-icon {
+      margin-right: 10px;
+    }
+  }
+}
```
MaverickW: I see a lot of custom CSS, despite the fact that you are using a CSS framework. Also, there are a lot of code duplications, for example `mat-icon { margin-right: 10px; }` included into multiple files.

src/app/core/nav-menu/components/license/license.component.ts
```diff
@@ -0,0 +1,34 @@
+import { ChangeDetectionStrategy, Component } from '@angular/core';
+
+@Component({
+  selector: 'cs-license',
+  changeDetection: ChangeDetectionStrategy.OnPush,
+  template: `
+    <div class="license">
+      {{ 'NAVIGATION_SIDEBAR.LICENSE.LICENSE_1' | translate: { year: getCurrentYear() } }}
+      <a href="https://bitworks.software/" target="_blank">
```
MaverickW: Links with `taget="_blank"` should have `rel="nofollow noopener"` because of security issues. Check this [article](https://habr.com/post/282880/) for details.

src/app/core/nav-menu/nav-menu.service.ts
tamazlykar: If you do so you will lose type below in pipe

src/app/core/nav-menu/components/app-nav/app-nav.component.html
tamazlykar: Because EventEmitter has a boolean type, I just need to send anything.

src/app/core/nav-menu/nav-menu.service.ts
tamazlykar: Okay, but it tested though `getSubroutes`

src/app/core/nav-menu/components/app-nav/app-nav.component.html
MaverickW: Is this the agreement, that you have to emit event right from the template? I don't really like when internal logic of events handling is included in the template. Imagine that one day you'll decide to change event emitter to Subject, in this case, you have also change the template.

We prefer to put such logic inside of public methods of the component, like `public open() { this.close.emit(true) }`. Also, it will allow you to check that `this.close` actually initiated and exists.

src/app/core/nav-menu/routes/accounts-subroutes.ts
```diff
@@ -0,0 +1,22 @@
...
+    path: '/events',
+    icon: 'mdi-calendar-text',
+    routeId: 'accounts',
+  },
+  {
+    text: 'NAVIGATION_SIDEBAR.SETTINGS',
+    path: '/settings',
+    icon: 'mdi-settings',
+    routeId: 'accounts',
+  }
```
MaverickW: Seems like `,` is missed here.

src/app/core/nav-menu/routes/app-nav-routes.ts
```diff
@@ -0,0 +1,20 @@
...
+    icon: 'mdi-cloud',
+    subroutes: virtualMachinesSubroutes,
+  },
+  {
+    id: 'accounts',
+    text: 'NAVIGATION_SIDEBAR.ACCOUNTS',
+    path: '/accounts',
+    icon: 'mdi-account-supervisor',
+    subroutes: accountsSubroutes,
+  }
```
MaverickW: Seems like `,` is missed here.

src/app/core/nav-menu/routes/app-nav-routes.ts
```diff
@@ -0,0 +1,20 @@
...
+import { virtualMachinesSubroutes } from './virtual-machines-subroutes';
+import { accountsSubroutes } from './accounts-subroutes';
+
+export const appNavRoutes: Route[] = [
+  {
+    id: 'virtual-machines',
+    text: 'NAVIGATION_SIDEBAR.VMS',
+    path: '/instances',
+    icon: 'mdi-cloud',
+    subroutes: virtualMachinesSubroutes,
```
MaverickW: Direct links to the other array can cause errors in case of mutation. Suggest change to `subroutes: [...virtualMachinesSubroutes]`

src/app/home/home.component.ts
```diff
@@ -15,14 +17,18 @@ import * as authActions from '../reducers/auth/redux/auth.actions';
 })
 export class HomeComponent extends WithUnsubscribe() implements OnInit {
   public disableSecurityGroups = false;
-  public isSidenavVisible$ = this.store.pipe(select(layoutSelectors.isSidenavVisible));
-  public allowReorderingSidenav$ = this.store.pipe(select(configSelectors.get('allowReorderingSidenav')));
+  public routes$: Observable<Route[]> = this.navRoutesService.getRoutes();
+  public currentRoute$: Observable<Route> = this.navRoutesService.getCurrentRoute();
+  public subroutes$: Observable<Subroute[]> = this.navRoutesService.getSubroutes();
```
MaverickW: I'm not quite sure that it should be initiated that way. You can find `this.navRoutesService` only in the constructor, and it's not so obvious.

Also, this initialization relies on code generation mechanisms which can be changed later by the developers of TypeScript. I suggest moving this logic into the constructor.

src/app/home/home.component.html
```diff
@@ -1,7 +1,12 @@
...
   </mat-sidenav>
 
-  <router-outlet></router-outlet>
+  <mat-sidenav-content>
+    <cs-section-nav [username]="username" [subroutes]="subroutes$ | async" (openAppNav)="sidenav.open()"></cs-section-nav>
+    <div class="primary-content">
+      <router-outlet></router-outlet>
+    </div>
+  </mat-sidenav-content>
 </mat-sidenav-container>
```
MaverickW: I have some concerns about these changes:

1. All logic related to exapnding\collapsing of menu items now encapsulated in the template and you can't check it using unit tests.
2. I don't really like the naming of `close` and `openAppNav`. They are using a different naming pattern: one of them includes an action target. It is not so obvious that the events should have specified target, it increases components coupling. For example `closeButtonClicked` sounds better and not overpromising.
3. By these changes, you missed the access to the sidenav state (opened, closed), but it can be helpful in the future.

src/app/core/nav-menu/nav-menu.service.ts
```diff
@@ -0,0 +1,58 @@
...
+      map(url => {
+        const matches = url.match(/^\/[A-Za-z-]*/);
+        return matches[0];
+      }),
+    );
+  }
+
+  private getAllSubroutes(): Subroute[] {
+    const subroutesArrays = appNavRoutes.map(route => route.subroutes);
+    return [].concat.apply([], subroutesArrays);
```
MaverickW: This construction a bit unreadable. Maybe it is better to use `reduce` instead?

```
appNavRoutes.map(route => route.subroutes)
  .reduce((x, acc) => [...acc, ...x], [])
```

src/app/core/nav-menu/nav-menu.service.ts
```diff
@@ -0,0 +1,58 @@
...
+      select(routerSelectors.getUrl),
+      filter(url => url != null),
+      map(url => {
+        const matches = url.match(/^\/[A-Za-z-]*/);
+        return matches[0];
+      }),
+    );
+  }
+
+  private getAllSubroutes(): Subroute[] {
```
MaverickW: What if subroutes can be changed on the flight in the future? Is there any problem to make them `Observable` at this moment?

src/app/core/nav-menu/nav-menu.service.ts
```diff
@@ -0,0 +1,58 @@
...
+        const subroutes = this.getAllSubroutes();
+        return subroutes.find(subroute => subroute.path === path)
+      })
+    );
+  }
+
+  private getCurrentSubroutePath(): Observable<string> {
+    return this.store.pipe(
+      select(routerSelectors.getUrl),
+      filter(url => url != null),
```
MaverickW: I suggest usage of strict comparison (`!==`) here. Also, maybe it is better to check that URL is a non-empty string?

src/app/core/nav-menu/nav-menu.service.ts
MaverickW: Also, you did something similar with `routes`

```
public getRoutes(): Observable<Route[]> {
    return of(appNavRoutes);
  }
```

src/app/core/nav-menu/nav-menu.service.ts
```diff
@@ -0,0 +1,58 @@
...
+  constructor(private store: Store<State>) {
+  }
+
+  public getRoutes(): Observable<Route[]> {
+    return of(appNavRoutes);
+  }
+
+  public getSubroutes(): Observable<Subroute[]> {
+    return this.getCurrentRoute().pipe(
+      map(route => route.subroutes)
```
MaverickW: Shouldn't you put such selectors in some kind of `index.ts` files? I saw something in other parts.

src/app/core/nav-menu/nav-menu.service.spec.ts
```diff
@@ -0,0 +1,45 @@
...
+      providers: [
+        NavMenuService,
+        { provide: Store, useClass: TestStore }
+      ]
+    });
+    service = TestBed.get(NavMenuService);
+    store = TestBed.get(Store);
+  });
+
+  it('should return an array of routes', () => {
```
MaverickW: As far as I know, tests with async tasks inside should be wrapped into `async` or you have to use `done` param. Seems like it is working only because of the nature of `Observable.of`, actually, it works synchronously. Just try to put a delay after the `Observable.of` and you will get trouble.

src/app/core/nav-menu/components/license/license.component.ts
tamazlykar: Style guide in progress #1061 
But we don't have any explicit rule for that. I did not add visibility to routes as this is an input property, also I didn't add visibility to output props.

src/app/core/nav-menu/components/license/license.component.ts
tamazlykar: No, we don't have any rule about this. I make this based on component size and simplicity.

src/app/core/nav-menu/components/app-nav/app-nav.component.ts
```diff
@@ -0,0 +1,18 @@
...
+  templateUrl: './app-nav.component.html',
+  styleUrls: ['./app-nav.component.scss']
+})
+export class AppNavComponent {
+  @Input()
+  routes: Route[];
+  @Input()
+  currentRoute: Route;
+  @Output()
+  close = new EventEmitter<boolean>();
```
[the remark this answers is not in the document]
tamazlykar: We don't have explicit conventions about output properties. I will add this to our local style guide.

src/app/core/nav-menu/routes/app-nav-routes.ts
tamazlykar: Is it will be helpful if we don't get a deep copy of the array?

src/app/core/nav-menu/nav-menu.service.ts
tamazlykar: Okay. But what do you think about extract `[].concat.apply([], subroutesArrays)` to a pure function with `flattenArray` name. Or better use a reduce approach. I just wondering for myself.

src/app/core/nav-menu/nav-menu.service.ts
tamazlykar: No this is not a problem. But how this will help us to implement  changes on the flight?

src/app/core/nav-menu/nav-menu.service.ts
tamazlykar: I don't understand you. Can you provide more information?
P.S. This is not a ngrx selector.

PR #1368, opened 2018-10-25
fix(disk-offering): enable change disk offering for volume

HeyRoach:


src/app/shared/actions/volume-actions/volume-resize/volume-resize.component.ts
```diff
@@ -78,8 +78,14 @@ export class VolumeResizeComponent implements OnInit, OnChanges {
...
   public resizeVolume(): void {
     const includeDiskOffering = this.diskOffering && !isRoot(this.volume);
-    const size = this.newSize ? { size: this.newSize } : {};
     const diskOffering = includeDiskOffering ? { diskofferingid: this.diskOffering.id } : {};
+
+    // send size only for custom offering and for same offering
+    const isResizableDiskOffering = isCustomized(this.diskOffering)
+      ? true
+      : diskOffering && this.volume.diskofferingid === diskOffering.diskofferingid;
+    const size = this.newSize && isResizableDiskOffering ? { size: this.newSize } : {};
```
wowshakhov: You can use `pickBy` to avoid this construct:
`includeDiskOffering ? { diskofferingid: this.diskOffering.id } : {}`.
For example, you do 
```
pickBy({
  fieldA: 1,
  fieldB: null,
  fieldC: undefined
}, Boolean)

And it filters all non null entries, resulting in 
```
{
  fieldA: 1
}
```
```

Also, these expressions are hard to understand, especially the nested ternary. I would like it to be a function.

In general, this parameter building function should not be placed inside a component. This should be done in an effect/selector.

src/app/shared/actions/volume-actions/volume-resize/volume-resize.component.ts
```diff
@@ -78,8 +78,9 @@ export class VolumeResizeComponent implements OnInit, OnChanges {
 
   public resizeVolume(): void {
     const includeDiskOffering = this.diskOffering && !isRoot(this.volume);
-    const size = this.newSize ? { size: this.newSize } : {};
     const diskOffering = includeDiskOffering ? { diskofferingid: this.diskOffering.id } : {};
+    const size = this.newSize && isCustomized(this.diskOffering) ? { size: this.newSize } : {};
```
wowshakhov: Move this to effect

PR #1373, opened 2018-10-26
fix(dialogs): formatting and layout

wowshakhov:


src/app/shared/components/vm-statistics/vm-statistics.component.ts
```diff
@@ -179,10 +179,14 @@ export class VmStatisticsComponent implements OnInit, OnChanges {
     return this.getStatsStringWithNoRestrictions(value, units, precision);
   }
 
-  public getStatsStringFor(resource: keyof ResourcesData, units?: string): Observable<string> {
+  public getStatsStringFor(
+    resource: keyof ResourcesData,
+    units?: string,
```
HeyRoach: an optional argument should be the last

wowshakhov: Ideally, this needs to be refactored with param objects. Will do this in a separate task later.

PR #1401, opened 2018-11-02
feat(vm-logs): settings

wowshakhov:


src/app/settings/settings.module.ts
```diff
@@ -1,10 +1,8 @@
 import { NgModule } from '@angular/core';
 import { CommonModule } from '@angular/common';
 import { ClipboardModule, ClipboardService } from 'ngx-clipboard';
-
```
tamazlykar: It's a good idea to separate third party imports and local imports. This described in Angular Style Guide

src/app/settings/settings.module.ts
```diff
@@ -15,6 +13,10 @@ import {
   ThemeSelectorComponent,
   VmSettingsComponent,
 } from './components';
+import { UpdateButtonFieldComponent } from './components/update-button-field/update-button-field.component';
+import { VmLogsMessagesComponent } from './components/vm-logs-settings/vm-logs-messages/vm-logs-messages.component';
```
tamazlykar: Why you did not add components in index.ts and did not export them as above?

src/app/vm-logs/redux/vm-logs.effects.ts
```diff
@@ -106,12 +107,13 @@ export class VmLogsEffects {
   @Effect()
   enableAutoUpdate$: Observable<Action> = this.actions$.pipe(
     ofType(vmLogsActions.VmLogsActionTypes.ENABLE_AUTO_UPDATE),
-    switchMap(() =>
-      timer(0, 10000).pipe(
+    withLatestFrom(this.store.pipe(select(UserTagsSelectors.getVmLogsShowLastMinutes))),
+    switchMap(([_, minutes]) =>
```
tamazlykar: Use implicit name instead of `_`

src/app/settings/settings.module.ts
wowshakhov: Add optimize imports sort by module

PR #1406, opened 2018-11-08
fix(resources): the number of free resources is not updated after committing actions to change resources

wowshakhov:
Previous PR: https://github.com/bwsw/cloudstack-ui/pull/1350


src/app/reducers/disk-offerings/redux/disk-offerings.reducers.ts
```diff
@@ -117,15 +117,30 @@ export const getAvailableOfferings = createSelector(
   },
 );
 
-export const isDiskOfferingAvailableByResources = (minSize: number) =>
+export const isDiskOfferingAvailableByResources = ({
+  diskOffering,
+  minSize,
```
tamazlykar: Rename property to clarify variable meaning,

PR #1431, opened 2018-11-23
fix(vm-creation): default security group

wowshakhov:


src/app/reducers/vm/redux/vm-creation.effects.ts
```diff
@@ -138,24 +138,17 @@ export class VirtualMachineCreationEffects {
   @Effect()
   vmSelectPredefinedSecurityGroups$: Observable<Action> = this.actions$.pipe(
     ofType(vmActions.VM_SECURITY_GROUPS_SELECT),
-    withLatestFrom(
+    switchMap(() =>
       this.store.pipe(
-        select(fromSecurityGroups.selectPredefinedSecurityGroups),
-        filter(groups => !!groups.length),
+        select(fromSecurityGroups.selectDefaultSecurityGroup),
+        skipWhile(value => !value || !value.id),
```
tamazlykar: Why do you use a skipWhile instead of a filter and check id?

PR #1447, opened 2018-11-28
feat(vm-logs): request frequency and requested interval

wowshakhov:


src/app/shared/models/config/config.interface.ts
```diff
@@ -22,6 +22,13 @@ export interface CustomizableConfig {
   sessionRefreshInterval: number;
   apiDocLink: string;
   extensions: ExtensionsConfig;
+  /*
+  *  Log View settings
+  * */
+  vmLogs: {
+    autoUpdateRefreshFrequency: number;
+    autoUpdateRequestedInterval: number;
```
tamazlykar: If you add new customizable properties, it will be great to add a description for this configuration in the config-guide

src/app/core/config/default-configuration.ts
```diff
@@ -41,6 +41,13 @@ export const customizableProperties: Readonly<CustomizableConfig> = {
     pulse: false,
     vmLogs: false,
   },
+  /*
+  * Log View settings
+  * */
+  vmLogs: {
+    autoUpdateRefreshFrequency: 1,
+    autoUpdateRequestedInterval: 30,
```
tamazlykar: According to the requirements, this value should be equal to 10.

PR #1450, opened 2018-11-29
fix(project): add repository url

wowshakhov:


package.json
```diff
@@ -2,6 +2,7 @@
   "name": "cloudstack-ui",
   "version": "1.411.25",
   "license": "Apache-2.0",
+  "repository": "github:bwsw/cloudstack-ui",
```
wowshakhov: conventional-changelog will automatically use repository URL if it is specified

PR #1459, opened 2018-12-04
fix(notifications): reset position

wowshakhov:


src/app/core/services/sidebar-width.service.ts
```diff
@@ -10,14 +10,18 @@ import { layoutStore, State, UserTagsSelectors } from '../../root-store';
  */
 @Injectable()
 export class SidebarWidthService {
-  public readonly width: Observable<number>;
-  private widthBehaviorSubject = new BehaviorSubject(0);
+  public readonly width$: Observable<number>;
+  public widthBehaviorSubject = new BehaviorSubject(0);
```
tamazlykar: This property should be private

PR #1475, opened 2018-12-07
fix(vm-creation): initial loading

wowshakhov:


src/app/reducers/vm/redux/vm.reducers.ts
```diff
@@ -43,6 +43,7 @@ export interface State extends EntityState<VirtualMachine> {
 
 export interface FormState {
   loading: boolean;
+  loaded: boolean;
```
tamazlykar: Do we really need this for a form? This looks like a derivative data from loading it this case.

src/app/vm/vm-creation/containers/vm-creation.container.ts
```diff
@@ -110,6 +115,7 @@ export class VmCreationContainerComponent implements OnInit {
   readonly account$ = this.store.pipe(select(fromAccounts.selectUserAccount));
   readonly zones$ = this.store.pipe(select(fromZones.selectAll));
   readonly sshKeyPairs$ = this.store.pipe(select(fromSshKeys.selectSshKeysForAccount));
+  readonly formInitSubscription: Subscription;
```
tamazlykar: Looks like unused

src/app/reducers/vm/redux/vm.reducers.ts
wowshakhov: If it's not loading it may be that it hasn't started loading yet.

PR #1483, opened 2018-12-10
fix(affinity-groups): don't show filters if there are no affinity groups

wowshakhov:


wowshakhov:
To test:
REQ_VM_DTLS_006

Issue #943, opened 2018-01-31
Manage limitation on reverting VM to VM snapshot when SO is changed


ershovaelena:
A part of the Epic #412

This issue covers the following requirements:
[REQ_SNAP_LIST_017] The system should limit the reverting VM to VM snapshot if the current SO differs from the one that VM had when VM snapshot was created

`VM snapshots which include both data volumes and memory can’t be kept if you change the VM’s service offering. Any existing VM snapshots of this type will be discarded.`

See [Apache Cloudstack official documentation](http://docs.cloudstack.apache.org/projects/cloudstack-administration/en/latest/virtual_machines.html?highlight=snapshot#virtual-machine-snapshots)

tamazlykar:
To test
REQ_SNAP_LIST_017

Nadya913:
In case if try to revert snapshot  to vm with changed SO with less cpus, cpu speed the error appears: Unexpected exception
Response:
{"queryasyncjobresultresponse":{"accountid":"162b26e6-052a-43bb-9116-acb7555b6d7d","userid":"f2c54886-ab40-44ed-9ee4-27dbacd47ceb","cmd":"org.apache.cloudstack.api.command.user.vmsnapshot.RevertToVMSnapshotCmd","jobstatus":2,"jobprocstatus":0,"jobresultcode":530,"jobresulttype":"object","jobresult":{"errorcode":530,"errortext":"Unexpected exception"},"created":"2018-12-17T14:01:04+0700","jobid":"8a2fb8da-704f-4e21-abf1-b7ccb5738e2f"}}

Steps:
1. Login and Create any VM with few CPU
2. Make a VM Snapshot
3. Change SO with less CPU, CPU speed
4. Revert snapshot to VM

Expected: dialog "Service offering differs from the original one" / "Вычислительное предложение отличается от первоначального"
Actual: unexpected error appears 
![error](https://user-images.githubusercontent.com/33111683/50071612-ef559980-0204-11e9-9b90-da065c834392.png)

PR #1495, opened 2018-12-14
feat(vm-snapshots): add validation for ability to revert a vm

tamazlykar:
Implements #943


src/app/root-store/server-data/vm-snapshots/vm-snapshots.effects.ts
```diff
@@ -211,8 +254,59 @@ export class VmSnapshotsEffects {
...
+    }),
+    exhaustMap(({ vmSnapshotId, vm }) => {
+      const params = {
+        resourceid: vmSnapshotId,
+        resourcetype: vmSnapshotEntityName,
+        key: vmSnapshotOfferingTagKey,
+      };
+      return this.tagService.getList(params).pipe(
+        /**
+         * if we unable to verify is revert allowed then we dispatch RevertAllowed in this case
```
wowshakhov: if we ARE unable

wowshakhov: * If we ARE unable to verify IF revert IS allowed, then we dispatch RevertAllowed
         * In this case, the user will get a server error if the revert is not allowed

src/i18n/en.json
```diff
@@ -60,7 +60,8 @@
     },
     "VM_SNAPSHOT": {
       "CONFIRM_DELETION": "Are you sure you want to delete this snapshot?",
-      "CONFIRM_REVERTING": "Are you sure you want to revert VM to this snapshot?"
+      "CONFIRM_REVERTING": "Are you sure you want to revert VM to this snapshot?",
+      "REVERT_NOT_ALLOWED": "Service offering differs from the original one"
```
wowshakhov: add period

src/i18n/ru.json
```diff
@@ -60,7 +60,8 @@
     },
     "VM_SNAPSHOT": {
       "CONFIRM_DELETION": "Вы уверены, что хотите удалить этот снимок?",
-      "CONFIRM_REVERTING": "Вы уверены, что хотите вернуть владеющую ВМ к состоянию этого снимка?"
+      "CONFIRM_REVERTING": "Вы уверены, что хотите вернуть владеющую ВМ к состоянию этого снимка?",
+      "REVERT_NOT_ALLOWED": "Вычислительное предложение отличается от первоначального"
```
wowshakhov: add period

PR #1506, opened 2018-12-27
fix(account-sidebar): fix account long name are above the other label…

HeyRoach:
…s in account sidebar


wowshakhov:
Add ellipsis:
selection_022

src/style/app.scss
```diff
@@ -167,6 +167,12 @@ mat-card.custom-card {
   }
 }
 
+.mat-content {
```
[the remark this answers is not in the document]
HeyRoach: I use ellipsis, but dots don't appear. I didn't catch why. Is it a problem?

src/app/account/account-sidebar/account-users/account-users.component.scss
```diff
@@ -0,0 +1,11 @@
+.mat-expansion-panel-header {
+  max-width: 100%;
```
wowshakhov: seems like it doesn't do anything

src/app/account/account-sidebar/account-users/account-users.component.scss
```diff
@@ -0,0 +1,11 @@
...
+}
+
+.entity-card-title {
+  overflow: hidden;
+  text-overflow: ellipsis;
+  white-space: nowrap;
+  max-width: 95%;
```
wowshakhov: is max-width: 95%; necessary?

src/app/account/account-sidebar/account-users/account-users.component.scss
```diff
@@ -0,0 +1,11 @@
+.mat-expansion-panel-header {
+  max-width: 100%;
+}
+
+.entity-card-title {
```
wowshakhov: you can replace entity-card-title with truncate from app.scss

src/app/account/account-sidebar/account-users/account-users.component.html
```diff
@@ -11,27 +9,20 @@
     (opened)="openItem(user)"
   >
     <mat-expansion-panel-header>
-      <mat-panel-title>
+      <mat-panel-title class="entity-card-title" [matTooltip]="getUserName(user)">
         {{ user.firstname }} {{ user.lastname }}
```
wowshakhov: why do you use a function call in one place and not in the other one? use literal strings in both places or the function call in both places

PR #1513, opened 2019-01-09
 feat(snapshot): add filtering by VMs to volume snapshots

HeyRoach:


src/app/snapshot/snapshots-page/snapshot-filter/snapshot-filter.container.ts
```diff
@@ -37,13 +36,15 @@ const FILTER_KEY = 'snapshotFilters';
       [firstDayOfWeek]="firstDayOfWeek$ | async"
       [selectedAccounts]="(filters$ | async).accounts"
       [selectedVms]="(filters$ | async).vmIds"
```
wowshakhov: You have `selectedVms` and `selectedVmIds` in your inputs. It may look like selectedVms means vm objects and selectedVmIds means vm ids, but you pass vmIds to the first input and volumeVmIds to the second input. This is confusing.

src/app/snapshot/snapshots-page/snapshot-filter/snapshot-filter.component.ts
```diff
@@ -55,6 +57,8 @@ export class SnapshotFilterComponent implements OnInit {
   public queryChange = new EventEmitter();
   @Output()
   public viewModeChange = new EventEmitter<SnapshotPageViewMode>();
+  @Output()
```
wowshakhov: I think we'd better refactor this component, so that VM view logic is separated from Volume view logic

src/app/snapshot/models/filters.model.ts
```diff
@@ -3,6 +3,7 @@ import { SnapshotType } from '../../shared/models';
 export interface Filters {
   accounts: string[];
   vmIds: string[];
+  volumeVmIds: string[];
```
wowshakhov: volumeVmIds is not entirely clear. Rename or add comment to describe what volumeVm means

HeyRoach:
rerun tests

PR #1520, opened 2019-01-11
feat(resource-limits): realm and env

wowshakhov:


src/app/core/config/default-configuration.ts
```diff
@@ -40,14 +40,19 @@ export const customizableProperties: Readonly<CustomizableConfig> = {
...
   /*
-  * Log View settings
-  * */
+   * Log View settings
+   * */
   vmLogs: {
     autoUpdateRefreshFrequency: 1,
     autoUpdateRequestedInterval: 30,
   },
+  resourceLimits: {
```
HeyRoach: remove resourceLimits from extensions

PR #1628, opened 2019-03-13
diskDetailsTest

evgeniiamokhova:
added disk details without tests of creation template and reverting disk to snapshot

### DIFF
--- a/e2e/pages/disk-list.po.ts
+++ b/e2e/pages/disk-list.po.ts
@@ -40,4 +40,21 @@ export class DiskList extends CloudstackUiPage {
   clickClose() {
     element(by.css('.backdrop.ng-star-inserted')).click();
   }
+
+  clickOnReadyDisk() {
+    element
+      .all(by.css('.entity-card.mat-card'))
+      .first()
+      .click();
+    const EC = browser.ExpectedConditions;
+    browser.wait(EC.presenceOf(element(by.css('.open'))), 5000, "Sidebar doesn't open");
+  }
+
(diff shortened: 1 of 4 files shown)


evgeniiamokhova:
run tests

e2e/pages/disk-list.po.ts
```diff
@@ -40,4 +40,21 @@ export class DiskList extends CloudstackUiPage {
   clickClose() {
     element(by.css('.backdrop.ng-star-inserted')).click();
   }
+
+  clickOnReadyDisk() {
+    element
+      .all(by.css('.entity-card.mat-card'))
+      .first()
```
rennervo: In this method, you click on the first available disk, but you must click on the disk in the status "Ready"

e2e/pages/disk-sidebar.po.ts
```diff
@@ -16,4 +21,88 @@ export class DiskSidebar extends CloudstackUiPage {
...
+      5000,
+      "Text area of disk description doesn't present",
+    );
+    element(by.name('textField')).clear();
+    element(by.name('textField')).sendKeys(description);
+    element
+      .all(by.css('.buttons button'))
+      .last()
+      .click();
+    element.all(by.css('.buttons button')).isPresent();
```
rennervo: Why do you need the last line of code in this function? Maybe it is superfluous?

evgeniiamokhova: Here I check the button "Cancel", but I incorrectly pointed to it

PR #1681, opened 2019-12-26
Smart HTTP VM Access Helper

### LINKED ISSUE #1680: Smart HTTP VM Access Helper implementation
### Description

Current HTTP VM Access Helper implementation generates the link as `http://<ip>:<port>/<path>`, where `<ip>` is taken from VM information, `<port>/<path>` is taken from VM tags, which are filled when VM created from a template.

The problem is that a user may click the helper link when VM is still in the booting state and VM HTTP software inside VM is not loaded yet, so user can get an error.

This feature proposal is to implement __smart__ HTTP Access Helper which relies on a small backend Node Service presented as an endpoint for UI which is requested by UI about the HTTP Status of `http://<ip>:<port>/<path>`. The backend Node implementation should use HEAD request to the URI and return OK if `200 OK` is received. This means that the link can be shown to the user in the UI.

The implementation must use configurable timeout (default 60 seconds) which is limit time to wait for a remote URI to reach 200 OK state. If the state not reached after the timeout, then UI must display timeout error notice instead of URI.

If the backend Node endpoint is not accessible, then UI must just show URI instantly.

Implement the feature and necessary tests.

andrewbents:
closes #1680


src/app/vm/vm-actions/vm-actions-component/vm-http-access-link/vm-http-access-link.component.ts
```diff
@@ -0,0 +1,77 @@
+import { Component, Input, OnDestroy, OnInit } from '@angular/core';
+import { concat, interval, Subscription } from 'rxjs';
+import { finalize, repeat, switchMap, take } from 'rxjs/operators';
+import { VirtualMachine } from '../../..';
+import { HttpAccessHelperService } from './http-access-helper.service';
+import { VmReachability } from './vm-reachability.enum';
+
+export const POLL_PERIOD = 5000;
```
wowshakhov: We can this to config json

src/app/vm/vm-actions/vm-actions-component/vm-http-access-link/vm-http-access-link.component.html
```diff
@@ -0,0 +1,24 @@
+<ng-container *ngIf="reachabilityState !== Reachability.Unknown" [ngSwitch]="reachabilityState">
+  <span *ngSwitchCase="Reachability.Unreachable">{{ 'HTTP_ACCESS.UNREACHABLE' | translate }}</span>
+  <a
+    *ngSwitchDefault
+    [href]="vm | vmHttpAddress"
+    class="action-link"
+    target="_blank"
+    rel="noopener"
+    >{{ vm | vmHttpAddress }}</a
```
wowshakhov: formatting

andrewbents: that's prettier

projects/http-access-helper/src/request.js
```diff
@@ -0,0 +1,18 @@
+const axios = require('axios');
+const { delay } = require('./util');
+
+function requestWithTimeout(url, timeout) {
+  const cancelToken = axios.CancelToken.source();
```
pm-Egor-Genning: maybe use axios.default.timeout instead of manual

src/app/vm/vm-actions/vm-actions-component/vm-http-access-link/http-access-helper.service.ts
```diff
@@ -0,0 +1,34 @@
...
+
+@Injectable({
+  providedIn: 'root',
+})
+export class HttpAccessHelperService {
+  constructor(private http: HttpClient, private httpAccess: HttpAccessService) {}
+
+  getReachibility(vm: VirtualMachine): Observable<VmReachability> {
+    const address = this.httpAccess.getAddress(vm);
+    return this.http
```
pm-Egor-Genning: not secure access to node `http-access-helper`

PR #1686, opened 2020-01-08
Login form enchancements

### LINKED ISSUE #1417: Login form enhancements
The following enhancements are proposed for the implementation.

1. Rework error notifications: show one notification for all cases of incorrectly entered data
2. Remove * (required fields) from username, password
3. Remove show/hide for a domain, always show domain
4. Add a throbber for login button. When logging in is proceeded, we will show that the system responds and processes the result.

To discuss

andrewbents:
Closes #1417 

login

### DIFF
--- a/src/app/auth/login.component.ts
+++ b/src/app/auth/login.component.ts
@@ -6,7 +6,6 @@ import { first } from 'rxjs/operators';
 
 import { AuthService } from '../shared/services/auth.service';
 import { SnackBarService } from '../core/services/';
-import { LocalStorageService } from '../shared/services/local-storage.service';
 import { configSelectors, State } from '../root-store';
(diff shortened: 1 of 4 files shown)


src/app/auth/login.component.ts
```diff
@@ -76,8 +68,11 @@ export class LoginComponent implements OnInit {
   }
 
   private login(username: string, password: string, domain: string): void {
+    this._loading = true;
+
     this.auth
       .login(username, password, domain)
+      .pipe(finalize(() => (this._loading = false)))
       .subscribe(() => this.handleLogin(), error => this.handleError(error));
```
Vnam92: can be moved into `pipe`

PR #1696, opened 2020-01-13
Add date range to snapshot filters

### LINKED ISSUE #1439: Implement start and end date filters in VM/Volume snapshots
There are 2 possible ways for the implementation
1. Start date - one year ago, end date - today by default
2. Both dates are empty by default 

Check Material components and propose PO. To discuss.

andrewbents:
closes #1439 

image


src/app/snapshot/store/snapshot-page.selectors.ts
```diff
@@ -74,6 +74,21 @@ export const getSelectedSnapshotForSidebar = createSelector(
   },
 );
 
+function isBetween(date: moment.MomentInput, a: Date | undefined, b: Date | undefined) {
```
Vnam92: add return type

